PR #4, opened 2018-12-29
Provide source files to fable compiler from nuget package.

enerqi:
Following the [Fable FAQ](https://fable.io/faq/#which-files-need-to-be-included-in-the-nuget-package-).
FsToolkit.ErrorHandling seems to work with fable - I tried some examples from the tests inside a Fable app.

### DIFF
--- a/src/FsToolkit.ErrorHandling/FsToolkit.ErrorHandling.fsproj
+++ b/src/FsToolkit.ErrorHandling/FsToolkit.ErrorHandling.fsproj
@@ -32,5 +32,11 @@
     <Compile Include="List.fs" />
     <None Include="Script.fsx" />
   </ItemGroup>
+
+  <!-- Add source files to "fable" folder in Nuget package -->
+  <ItemGroup>
+    <Content Include="*.fsproj; **\*.fs" PackagePath="fable\" />
+  </ItemGroup>
+
   <Import Project="..\..\.paket\Paket.Restore.targets" />
 </Project>


demystifyfp:
Thanks, @enerqi 

I haven't tried this with Fable. Let me publish a NuGet version with this change. Can you verify the new NuGet package?

enerqi:
I'm finding that there is some sort of packaging problem - I've not learnt enough about .NET packaging and assemblies to know the correct solution. When I open the `FsToolkit.ErrorHandling.0.0.10.nupkg` file (they are just zip files) I see that there is no `fable` folder. 

So, it's something to do with the Fake build. The Fake "NuGet" target is creating a nupkg without the fable folder. On the other hand if I directly try `dotnet pack --output %HOME%\dev\fsharp\nupkgs -c Release src\FsToolkit.ErrorHandling\FsToolkit.ErrorHandling.fsproj` then I get a nupkg labelled version 1.0.0 that has the correct files in it.

Looks like some other open source F# projects are using `DotNetCli.Pack` instead of `Paket.pack`.

enerqi:
I see that `dotnet pack` additionally adds the `lib/net461/FsToolkit.Errorhandling.dll` inside the nupkg. The NuGet Fake target only puts in `lib/netstandard2.0/FsToolkit.Errorhandling.dll`.  Otherwise the outputs look similar.

demystifyfp:
Hi @enerqi. I guess the difference is because the build script uses Paket to create the NuGet package. It targets both .NET Framework and DotNet core. But I am not sure how to fix this. 

Hi @alfonsogarciacaro, @forki & @ncave,
Can you folks help us here in making FsToolkit.ErrorHandling to work with Fable?

alfonsogarciacaro:
Is it working after #5? This PR looks good to me. I usually use the `dotnet pack` command to create the packages.

demystifyfp:
Hi @alfonsogarciacaro. Yes, it is working after #5 . Thanks :)

PR #5, opened 2019-01-02
Dotnet pack fable files

enerqi:
Using the latest Fake libs so can use the relevant Fake.Dotnet features. Judging by the [dotnet pack](https://docs.microsoft.com/en-us/dotnet/core/tools/dotnet-pack?tabs=netcore2x) and [nuget metadata](https://docs.microsoft.com/en-us/dotnet/core/tools/csproj#nuget-metadata-properties) docs I hope this is correct.


demystifyfp:
I already published it to NuGet (5 hours ago). But it is not reflecting there! Do you have any idea?

demystifyfp:
@enerqi I guess it could be because of the `bin` folder location

```fsharp
Target.create "NuGet" (fun _ ->
    let ReleaseNotes = String.toLines release.Notes
    "src/FsToolkit.ErrorHandling"
    |> DotNet.pack(fun p ->
           { p with
               OutputPath = Some "bin"
               Configuration = DotNet.BuildConfiguration.Release
               MSBuildParams = {MSBuild.CliArguments.Create() with
                                    // "/p" (property) arguments to MSBuild.exe
                                    Properties = [("Version", release.NugetVersion)
                                                  ("PackageReleaseNotes", ReleaseNotes)]}})
)

Target.create "PublishNuget" (fun _ ->
    Paket.push(fun p ->
        { p with
            PublishUrl = "https://www.nuget.org"
            WorkingDir = "bin" })
)
```

Can you verify this as well ?

enerqi:
Yes, it seems that OutputPath is relative to the directory passed to dotnet pack. 
`OutputPath = Some "../../bin"` works. I guess the other approach would be to set the `WorkingDir` on the "PublishNuget" target to `src/FsToolkit.ErrorHandling` and leave the pack OutputPath as `bin`.

I tried passing "." to dotnet pack (starting from the project root dir) but that packages up tests aswell and puts the nupkgs under the project directories, so that doesn't look any prettier.

PR #6, opened 2019-01-03
Match dotnet pack output path to the path expected by PublishNuget

enerqi:


demystifyfp:
Thanks, @enerqi Can you verify the version `0.0.12`? 

 It'd be helpful to the community if you can write short documentation on how to use FsToolkit.ErrorHandling in a Fable Project. 

https://github.com/demystifyfp/FsToolkit.ErrorHandling/tree/master/gitbook

enerqi:
Yes, that's working from nuget.org now. Yeah, there's probably something useful to document after using it in a real fable project. I think it'll just work except for some async subleties that Fable may not support.

PR #20, opened 2019-04-23
Generate symbol package

cmeeren:
This PR attempts to add SourceLink support by generating a symbol package. Note that:

* There may or may not be CI or build.fsx changes needed to actually publish them (I don't know exactly how `Paket.push` works with regards to symbols; for Cvdm.ErrorHandling, I could only get it to work by using the publish mechanisms of AppVeyor, but things might have changed)
* I removed `<DebugType>portable></DebugType>` because [it's the default for .NET SDK projects](https://github.com/dotnet/sourcelink)


cmeeren:
Having some problems with rebasing, will notify when it's fixed

PR #21, opened 2019-04-23
Generate symbol package

cmeeren:
This PR attempts to add SourceLink support by generating a symbol package. Note that:

* There may or may not be CI or build.fsx changes needed to actually publish them (I don't know exactly how `Paket.push` works with regards to symbols; for Cvdm.ErrorHandling, I could only get it to work by using the publish mechanisms of AppVeyor, but things might have changed)
* I removed `<DebugType>portable</DebugType>` because [it's the default for .NET SDK projects](https://github.com/dotnet/sourcelink)


demystifyfp:
Thanks, @cmeeren for cleaning and updating the documents. 

Is it a WIP PR?

cmeeren:
No, this is as final as I can make it. Whether the CI or build script actually publishes symbols I can't say. You'll have to test on the next release, I think. (I don't even know how to check it.)

TheAngryByrd:
I haven't used the new Microsoft sourcelink stuff but https://github.com/ctaggart/SourceLink#test might still work as a good way to verify it's working

demystifyfp:
@cmeeren I don't have much idea about it as well. If you are sure, then I can merge it and publish a new version.

cmeeren:
What I am sure about (and have tested) is that this PR makes sure a symbol nupgk is generated.

What I am not sure about, and have no way to find out, is how the publish mechanism interacts with the symbol package.

When using AppVeyor to upload the nupkg files, everything seems to work fine. When I originally tried doing it using a FAKE build task, I had problems where it didn't really understand what a symbol file was, so it ended up uploading the symbol file first as the main package, and then the upload of the main package failed because it was already published. This was no crisis - the symbol package includes the DLLs and docs just as the main package does - but it was what made me change from FAKE publishing to AppVeyor publishing.

cmeeren:
By the way, AFAIK an alternative to a separate symbol package is to include the PDB in the main package. I have done that too for other projects.

demystifyfp:
@cmeeren Let's verify its behaviour in the next release

cmeeren:
I see that `<DebugType>portable</DebugType>` got added again in the merge. It probably doesn't hurt, but it's not needed because [it's the default](https://github.com/dotnet/sourcelink).

PR #46, opened 2019-05-10
Jobresult

### LINKED ISSUE #17: Making TaskResult and JobResult libraries
It was requested [via twitter](https://twitter.com/sergey_tihon/status/1116247497234628613) to have a `TaskResult` added to this library.  I do believe this would be a good idea, as I also work with task heavy CE workloads and rely on the Result type.  

As mentioned in that tweet thread, since the main library wants to be Fable compatible, it would have to be its own separate library. `FsToolkit.ErrorHandling.Tasks` (something something naming is hard).

In addition to this, I use Hopac quite a bit and have authored the `Chessie.Hopac`. Similar to the above, it would be its own library `FsToolkit.ErrorHandling.Hopac`.  

I'd be willing to implement both of these libraries if this proposal makes sense.

TheAngryByrd:
Closes #17 with Adding `Job<Result<_,_>`


demystifyfp:
Hi @TheAngryByrd, I am currently held up. Will merge the changes over this weekend.

demystifyfp:
@TheAngryByrd I am getting this weird build error on my machine, and because of it, I am not able to publish the package. In CI also, it is [failing](https://travis-ci.org/demystifyfp/FsToolkit.ErrorHandling/builds/535817560)!

```
~/FsToolkit.ErrorHandling/src/FsToolkit.ErrorHandling.JobResult/JobResultCE.fs(157,15): error FS0074: The type referenced through 'System.OperationCanceledException' is defined in an assembly that is not referenced. You must add a reference to assembly 'System.Threading.Tasks'. [~/FsToolkit.ErrorHandling/src/FsToolkit.ErrorHandling.JobResult/FsToolkit.ErrorHandling.JobResult.fsproj]
```

I tried adding this package, but it didn't succeed. 

Any idea?

TheAngryByrd:
@demystifyfp https://github.com/demystifyfp/FsToolkit.ErrorHandling/pull/47 should fix it

PR #47, opened 2019-05-22
Adds FrameworkPathOverride for mono

TheAngryByrd:
To use the dotnet sdk to build full framework target frameworks on mac/linux you need to teach the dotnet sdk where the Full Framework binaries are on the machine by setting FrameworkPathOverride


tamizhvendan:
Thanks @TheAngryByrd

PR #50, opened 2019-05-29
Add Result.sequenceAsync

cmeeren:
Fixes #50 

I haven't added any tests because I can't think of anything to test that the compiler does not guarantee. AFAIK you'd have to use reflection in order to have a bug.

I also haven't added any doc comments to the function since the other `sequence` functions don't have that.

### DIFF
--- a/.paket/Paket.Restore.targets
+++ b/.paket/Paket.Restore.targets
@@ -252,7 +252,7 @@
     </PropertyGroup>
 
     <ItemGroup>
-      <_NuspecFiles Include="$(AdjustedNuspecOutputPath)\*.nuspec"/>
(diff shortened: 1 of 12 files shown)


TheAngryByrd:
The one test I can think of is it gives you the result you've passed in. Theoretically this function could just return unchecked defaultOf. I know it would be stupid but at least we have a real test.

TheAngryByrd:
Could you also add implementations for Task and Job libraries?

tamizhvendan:
Thanks, @cmeeren. Can you add the documentation as well?

tamizhvendan:
Thanks, @cmeeren

PR #51, opened 2019-05-30
Fable tests

### LINKED ISSUE #49: Add Fable Tests
Since this library claims to support Fable, we should run tests against javascript to prove that it does.

https://github.com/Zaid-Ajaj/Fable.Mocha has an expecto like API that allows us to use compiler switch to easy switch between dotnet or javascript runtimes.

I've already started down this and will send a PR about this soon.

TheAngryByrd:
Closes #49 

This allows us to run our main library on node using Fable.Mocha.  The api for Fable.Mocha is the same as Expecto and allows us to transpile out tests to run on Mocha with the help of a compiler directive. 

Unfortunately there's not way to do the reflection magic expecto does, so we have to specify all the tests we want to run.


tamizhvendan:
Thanks, @TheAngryByrd. It will be handy. 
Can you fix the build as well?. It is failing because of some dotnet core version.

TheAngryByrd:
Should be good to go now

tamizhvendan:
Thanks @TheAngryByrd

Issue #55, opened 2019-07-20
Consider adding Result.requireNonNull


danyx23:
I really like this library! I just found myself wanting a function that takes a nullable value and makes sure that it is not null by turning it into an Error value if it is null (useful for the many dotnet function that might return null). Would you consider a PR that adds something like the code below?

```fsharp
let requireNotNull (errorValue : 'error) (value : 'a when 'a : null ) =
    match value with
    | null -> Error errorValue
    | nonnull -> Ok nonnull
```

cmeeren:
Would `Option.ofObj >> Result.requireSome` suit your purposes?

demystifyfp:
@cmeeren It'd help. But I feel @danyx23 requirement makes sense in the context of other such helper functions. What do you recommend? 

One suggestion on the name of the function, can we have it as `requireNonNull`?

cmeeren:
The helper function makes sense to me. 🙂

I'd call it requireNotNull (as OP suggested) instead of requireNonNull though, to fit with other functions (e.g. notEqualTo) (and in my mind it sounds better), but I certainly don't want to create a fuss over the difference.

danyx23:
Yes I think while `Option.ofObj >> Result.requireSome` is also a nice way it would be good to have something specific for this rather common case that is a bit more obvious when reading the code. 

I am totally fine with requireNotNull. I'll send a PR and then we can discuss any further details there. Thansk!

danyx23:
Sent a PR. Let me know what you think. I only realized in the PR that there are some whitespace changes because my editor is set to remove trailing whitespace and doesn't show whitespace diffs when staging changes - hope you don't mind.

demystifyfp:
Thanks, @danyx23 for the PR. I've merged the changes and published the new version `1.2.3`. Can you verify the changes and close this issue?

danyx23:
Looks good. I have some strange issue with the update of the nuget file in the project that I am using this but the assembly has the new code and everything should work so I'm closing the issue.

PR #56, opened 2019-07-21
Add Result.requireNotNull function, tests and docs

danyx23:
This implements the requireNotNull function discussed in #55.

### DIFF
--- a/gitbook/result/others.md
+++ b/gitbook/result/others.md
@@ -26,6 +26,13 @@ Converts an Option to a Result, using the given error if Some.
 ```fsharp
 'a -> 'b option -> Result<unit, 'a>
 ```
+### requireNotNull
+
+Converts a nullable value to a Result, using the given error if null.
+```fsharp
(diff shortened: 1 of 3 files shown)


gitbook/result/others.md
```diff
@@ -26,6 +26,13 @@ Converts an Option to a Result, using the given error if Some.
 ```fsharp
 'a -> 'b option -> Result<unit, 'a>
 ```
+### requireNotNull
+
+Converts a nullable value to a Result, using the given error if null.
+```fsharp
+'a -> 'b : null -> Result<'b, 'a>
```
danyx23: I'm not sure what the correct annotation syntax is here for the generic constraint `null`. Do you know?

cmeeren: IMHO it's not necessary to specify constraints like that in the docs.

demystifyfp:
@danyx23 Other than @cmeeren comment, I don't have any other comments. Thanks for the PR :)

PR #59, opened 2019-10-17
Improve result CE overload resolution

cmeeren:
The `result` CE has problems with overload resolution due to overloading with `Choice`.

For example, when defining a function that accepts a `Result` value (or a `Result`-returning function), I would expect to be able to bind/return this value even when not explicitly annotating the parameter.

An example of such a function (trivial, but I'm sure you can extrapolate to more useful ones):

```f#
let myFun getSomeRes =
  result {
    return! getSomeRes "foo" "bar" 42
  }
```

This fails, and currently I would need to write

```f#
let myFun (getSomeRes: _ -> _ -> _ -> Result<_, _> =
```

That shouldn't be necessary. This PR fixes that by moving the `Choice` members so that they are extension members.

By the way, the added tests are purely compile-time tests.

### DIFF
--- a/src/FsToolkit.ErrorHandling/ResultCE.fs
+++ b/src/FsToolkit.ErrorHandling/ResultCE.fs
@@ -12,9 +12,6 @@ module ResultCE =
     member __.ReturnFrom (result: Result<'T, 'TError>) : Result<'T, 'TError> =
       result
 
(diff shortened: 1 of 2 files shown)


TheAngryByrd:
lgtm

demystifyfp:
Thanks @cmeeren Pushed the latest version.

PR #63, opened 2019-12-19
Add Result.valueOr

cmeeren:
I think this is useful. Feel free to disagree. :)

### DIFF
--- a/src/FsToolkit.ErrorHandling/Result.fs
+++ b/src/FsToolkit.ErrorHandling/Result.fs
@@ -165,3 +165,9 @@ module Result =
           return Ok x
       | Error err -> return Error err
     }
+    
+  /// Returns the Ok value or runs the specified function over the error value.
+  let valueOr f res =
+    match res with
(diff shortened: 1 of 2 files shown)


cmeeren:
Hmm. We could, but I'm not sure it makes perfect sense for a function called `defaultWith` to accept the error case. Also `Option.defaultWith` accepts a thunk; I'm not sure whether thunks semantically should always accept `unit` or if they can accept other arguments.

FSharpPlus calls this function `valueOr` (for their `Validation` type).

All of the above may be non-issues, however. I'll let you decide. :)

TheAngryByrd:
Sounds good to me. 

Then I’d say just add two tests as well and let’s the this merged.

cmeeren:
There. Edited directly through GH, hope there's no silly mistakes.

demystifyfp:
Thanks, @cmeeren for the new function. Would you be able to add it in the documentation as well?

PR #69, opened 2020-02-28
WIP: Implement Seq module

joemphilips:
In my project, I want to apply sequenceResultA for, `Array<Result<_,_>>`
But converting it to `List` will probably hurt performance.
So I want to have more generic ` Seq.sequenceResultA` and such.

Most of the code is simply copied from `List.fs` and tweaked.

Tell me if I'm doing something wrong.
* [x] code.
* [x] test
* [ ] performance comparison to same functions in `List.fs` by `Expecto.Performance`

### DIFF
--- a/tests/FsToolkit.ErrorHandling.Tests/SeqPerformanceTests.fs
+++ b/tests/FsToolkit.ErrorHandling.Tests/SeqPerformanceTests.fs
@@ -1,0 +1,42 @@
+module SeqPerformanceTests
+
+#if !FABLE_COMPILER
+open Expecto
+open Expecto.Tests
+open Expecto.Performance
+open FsToolkit.ErrorHandling
+open SampleDomain
+
+let inline repeat n a arg =
+    [for _ in 0..(n - 1) -> a arg]
+
+let inline repeat10 a arg =
+    repeat 10 a arg
+
+let traverseResultTests =
+    testSequenced <| testList "Seq performance tests " [
+        testCase "applying Seq.sequenceResultM against array must be faster than using List module" <| fun _ ->
+            let arr = [| for i in 1..127 -> Ok (i)|]
+            Expect.isFasterThan
+                (fun () -> repeat10 (Seq.sequenceResultM >> Result.map(Seq.toArray)) arr )
+                (fun () -> repeat10 (Array.toList >> List.sequenceResultM >> Result.map(List.toArray)) arr)
+                ""
+        testCase "applying Seq.sequenceResultA against array must be faster than using List module" <| fun _ ->
+            let arr = [| for i in 1..127 -> Ok (i)|]
+            Expect.isFasterThan
+                (fun () -> repeat10 (Seq.sequenceResultA >> Result.map(Seq.toArray)) arr )
+                (fun () -> repeat10 (Array.toList >> List.sequenceResultA >> Result.map(List.toArray)) arr)
+                ""
+        testCase "applying Seq.sequenceResultA against array must be faster than using List module (reference type)" <| fun _ ->
+            let arr = [| for i in 1..127 -> Tweet.TryCreate(sprintf "%d" i)|]
+            Expect.isFasterThan
+                (fun () -> repeat10 (Seq.sequenceResultA >> Result.map(Seq.toArray)) arr )
+                (fun () -> repeat10 (Array.toList >> List.sequenceResultA >> Result.map(List.toArray)) arr)
+                ""
+    ]
+
+let allTests = testList "Seq performance tests" [
+    traverseResultTests
+]
+
+#endif
--- a/package.json
+++ b/package.json
@@ -1,8 +1,8 @@
 {
   "devDependencies": {
-    "fable-compiler": "^2.4.6",
+    "fable-compiler": "2.4.19",
     "fable-splitter": "^2.1.10",
-    "mocha": "^6.1.4"
+    "mocha": "^7.1.0"
   },
   "scripts": {
     "pretest": "fable-splitter tests/FsToolkit.ErrorHandling.Tests -o dist/tests --commonjs",
--- a/src/FsToolkit.ErrorHandling/FsToolkit.ErrorHandling.fsproj
+++ b/src/FsToolkit.ErrorHandling/FsToolkit.ErrorHandling.fsproj
@@ -37,6 +37,7 @@
     <Compile Include="OptionCE.fs" />
     <Compile Include="AsyncOptionCE.fs" />
     <Compile Include="List.fs" />
+    <Compile Include="Seq.fs" />
     <None Include="Script.fsx" />
   </ItemGroup>
 
--- a/src/FsToolkit.ErrorHandling/Seq.fs
+++ b/src/FsToolkit.ErrorHandling/Seq.fs
@@ -1,0 +1,113 @@
+namespace FsToolkit.ErrorHandling
+
+[<RequireQualifiedAccess>]
+module Seq =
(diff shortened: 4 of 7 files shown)


joemphilips:
conversion from list to seq will result to empty seq sometimes in Fable, I don't know why. I'm not a Fable expert.
I stopped supporting `Seq` in Fable, I think performance is rarely an issue when using fable anyway, so this should be fine.
I will continue if I got approval.

TheAngryByrd:
> conversion from list to seq will result to empty seq sometimes in Fable, I don't know why. I'm not a Fable expert.

Hmm that is bizarre indeed.  Does this sound like a bug @alfonsogarciacaro @MangelMaxime @Zaid-Ajaj?

MangelMaxime:
@TheAngryByrd Perhaps there is a bug but I don't know how to reproduce it.

I tested some dummy code like:

```fs
open Fable.Core

let s = seq [ "m"; "" ]


let l = s |> Seq.toList

JS.console.log(s)
JS.console.log(l)
```

and it's working as expected.

joemphilips:
OK I will make clear what is exactly a problem.

Zaid-Ajaj:
A good idea would be to update Fable to latest stable version to make sure it isn't something that was already fixed

joemphilips:
Ok it seems that the problem is, `Seq.tail` and `Seq.skip` is not working in fable.

```fs
open Fable.Core
let tweets = [ "Hi"; "Hello"; "Hola" ] |> List.toSeq
JS.console.log(sprintf "tweets are %A" (JS.JSON.stringify(tweets))
JS.console.log(sprintf "tweet head is %A" (JS.JSON.stringify(tweets |> Seq.head)))
JS.console.log(sprintf "tweet tail is %A" (JS.JSON.stringify(tweets |> Seq.tail)))
JS.console.log(sprintf "tweet skip 1 is %A" (JS.JSON.stringify(tweets |> Seq.skip(1)))
```
returns

```
tweets are ["Hi","Hello","Hola"]
tweet head is "Hi"
tweet tail is {}
tweet skip 1 is {}
```

I've been looking into how to update fable compiler, or at least to check the version.
(Again, I am new to Fable)

joemphilips:
To make myself clear, I still think supporting `Seq` in fable is unnecessary, at least in this PR.

joemphilips:
I updated `fable-compiler` by  `yarn upgrade`, but version did not change and the result was the same.

MangelMaxime:
I opened an issue because even if you don't need seq in this PR. We do have a bug in Fable Seq module :)

In order to update `fable-compiler` you indeed need to run `yarn upgrade` in general I do `yarn upgrade --latest`.

joemphilips:
I have done `yarn upgrade --latest` but `fable-compiler` and `fable-splitter` version in `package.json` does not change, the version is

```
 "fable-compiler": "^2.4.6",
 "fable-splitter": "^2.1.10",
```
according to `package-lock.json` , `fable-compiler` version is `2.4.6` .

I also manually changed `fable-compiler` to version `2.4.16` (which is the latest release).
But the problem continues.

So this is clearly the problem of fable compiler side, I suppose we should continue the discussion on the issue @MangelMaxime   have just opened.

FYI my node and yarn version is 

```
$ node --version
v12.13.1
$ yarn --version
1.21.0
```

MangelMaxime:
@joemphilips If you are using `yarn` you should have a `yarn.lock` and not a `package-json.lock`.

The issue has been checked on Fable side what you reported is actually the expected behavior. Indeed,  `seq` are lazy so when you applying `JSON.stringify` to it, it's printing an empty object. 

In latest version of Fable (2.4.17) you can now use `sprintf "%A"` or ``printf "%A"`` against a `seq` value and it will print a "JSON" representation of the object and not just `Object [Object]`.

joemphilips:
@MangelMaxime  Thanks for the response. sorry for my ignorance about yarn packaging and behavior of fable

I updated the fable version as you mentioned and checked that it is working correctly, there is no problem about printing seq.

After a bit of a further research, I realized that the problem I was facing here regards to the strange behavior of `Seq.tail` .

```fsharp
      let tweets = seq [ yield "Hi"; yield "Hello"; yield "Hola" ]

      JS.console.log(sprintf "tweets are %A" tweets)
      let h = tweets |> Seq.head
      let t = tweets |> Seq.tail
      JS.console.log(sprintf "tweet head is %A" h)
      JS.console.log(sprintf "tweet tails are %A" t)

      let tweetsR = Ok (tweets)
      JS.console.log(sprintf " tweetR is %A" (tweetsR))
      match tweetsR with
      | Ok ys ->
          JS.console.log(sprintf "ys %A" (ys))

      let tweetsTailR = Ok (tweets |> Seq.tail)
      JS.console.log(sprintf " tweetsTailR is %A" (tweetsTailR))
      match tweetsTailR with
      | Ok ys ->
          JS.console.log(sprintf "ys %A" (ys))

      let tweetsSkipR = Ok (tweets |> Seq.skip 1)
      JS.console.log(sprintf " tweetsSkipR is %A" (tweetsSkipR))
      match tweetsSkipR with
      | Ok ys ->
          JS.console.log(sprintf "ys %A" (ys))
```
This prints
```
tweets are [Hi; Hello; Hola]                 // prints as expected
tweet head is Hi                                   // Seq.head also works as expected.
tweet tails are seq [Hello; Hola]         // Seq.tail works fine too.
 tweetR is Ok [Hi; Hello; Hola]            // lifting the seq to Result
ys [Hi; Hello; Hola]                              // does not change seq inside.
 tweetsTailR is Ok seq [Hello; Hola]   // but if we lift it to Result  *after* applying `Seq.tail`,
ys seq []                                              // it will result to empty seq.
 tweetsSkipR is Ok seq [Hello; Hola] // But Seq.skip seems to work fine.
ys seq [Hello; Hola]
```


So I decoded to continue by using `Seq.skip 1` instead of `Seq.tail` , which works fine.

tests/FsToolkit.ErrorHandling.Tests/SeqPerformanceTests.fs
```diff
@@ -0,0 +1,42 @@
...
+open SampleDomain
+
+let inline repeat n a arg =
+    [for _ in 0..(n - 1) -> a arg]
+
+let inline repeat10 a arg =
+    repeat 10 a arg
+
+let traverseResultTests =
+    testSequenced <| testList "Seq performance tests " [
```
joemphilips: I thought these tests will show that converting `Array -> Seq -> (Traversable operation against the seq) -> Array` will be much faster than `Array -> List -> (Traversable opreation against the List) -> Array` .

But interestingly the result was actually opposite. Latter was about 60x faster.

joemphilips:
I found out that `Seq.traverseResult[M|A]` function does not improve the performance against Array. So this PR is useless.
I suppose applying functional operation against an array was not a good idea in the first place.

MangelMaxime:
>  Thanks for the response. sorry for my ignorance about yarn packaging and behavior of fable

No problem :) We all started one day, and it's never easy when you are missing the ecosystems :)

Thank you for identifying a new bug. I will open an issue about it.

joemphilips:
> Thank you for identifying a new bug. I will open an issue about it.

@MangelMaxime 
Oh please don't. I already opened in here https://github.com/fable-compiler/Fable/issues/1996 . and I solved it when I cleaned the the whole cache and restarted.

MangelMaxime:
@joemphilips Ahah you were quicker than my edit :D

I saw the issue after my message.

joemphilips:
@MangelMaxime 
Thanks for making things simple. And sorry for another confusion ... 😅

PR #72, opened 2020-03-23
Use dotnet local tools

TheAngryByrd:
This makes use of the new [local tools](https://docs.microsoft.com/en-us/dotnet/core/tools/global-tools) that dotnet core 3.0 exposed.

paket and fake are now local tools to the repo, replacing the full framework version of paket and the global version of fake-cli.

This builds on top of #71


TheAngryByrd:
@demystifyfp I'd definitely for you to be aware of these changes before I merge them.

PR #97, opened 2020-08-15
TaskResult to Ply

NinoFloris:
Port of TaskResult to Ply

- Moves all CEs from Taskbuilder to Ply
- Inlines a lot more delegation methods, better for perf and stack traces.
- Implements loops on top of `uply` which is as cheap as it can get.
- Removes the net461 TFM as Ply does not support it.
- Pins FSharp.Core to .Net Core 2.1 LTS implicit reference version.

There's one unresolved matter, context insenstive binds. Ply does not support them out of the box (never had a need for them on .net core) so I have replaced them with normal binds to get this compiling. We could leave it, implement them here, or I could push a new version of Ply that has them; I don't have any strong preference.


FoggyFinder:
Just curious: Why? What advantages Ply gives over TaskBuilder ?

TheAngryByrd:
Thanks a ton for this @NinoFloris! 


> Inlines a lot more delegation methods, better for perf and stack traces.

❤️ 

> Implements loops on top of uply which is as cheap as it can get.

❤️ 

> Removes the net461 TFM as Ply does not support it

This one might really hurt. This has been the main reason I haven't ported to Ply yet.  Someday I'll want to give up netfx, but I'm unsure if I'd want to do that today.


> There's one unresolved matter, context insenstive binds. Ply does not support them out of the box (never had a need for them on .net core) so I have replaced them with normal binds to get this compiling

Yeah I could this this affecting people as well, most likely anyone dealing with UI work.  I'm unaware of anyone using it like this currently, but right after releasing, I know I'll get a bunch of issues about it 😅 .


--- 

Given those breaking changes, I don't think I can accept it as is. However, I'd be open to making a separate project/package (FsToolkit.ErrorHandling.PlyResult or something) that targets just Ply (Big old copy paste of what you have).  It allows people to try it out without breaking their current projects today in non obvious ways.  I'll make it heavily known in the release notes that we're planning on moving to Ply so please try it out, but in reality I know a subset won't and it'll be a surprise to them.

NinoFloris:
>Yeah I could this this affecting people as well, most likely anyone dealing with UI work. I'm unaware of anyone using it like this currently, but right after releasing, I know I'll get a bunch of issues about it 😅 .

Same, considering we want to get a bunch of libraries on board with Ply (which really just need `ConfigureAwait(false)`) I'll integrate it into Ply sometime soon.

>However, I'd be open to making a separate project/package (FsToolkit.ErrorHandling.PlyResult or something) that targets just Ply (Big old copy paste of what you have) 

I was thinking the same as a backup so I'm happy to hear you suggest it :)
Initially I had a line in my opening comment about releasing and how it would at the very least require a major version bump, that route could also be an option if I get the context insensitive things sorted. 

Generally speaking it's as described in the Giraffe issue, a change of namespaces, one bulk search and replace should do it.

NinoFloris:
Ah and on the netfx tfm, ns2.0 should be properly supported under net472 right?

baronfel:
Yes, net472 should be totally compatible with ns2.0 (per the infamous compatibility chart).

I would honestly be very supportive of dropping the net461 TFM. We're about to do so in FSharp.Compiler.Services, for example, and request that folks use net472 at minimum, so we can ship a netstandard-only dll. It's just _so much less_ complexity, and net472 is positively ancient at this point (more than 2 years old)!

TheAngryByrd:
> Ah and on the netfx tfm, ns2.0 should be properly supported under net472 right?

Yeah I think this is true. 

> It's just so much less complexity, and net472 is positively ancient at this point (more than 2 years old)!

I think we forget the pace as which Windows developers projects tend to move 😢 . 2 years is pretty new on their timescale. 

My personal opinion about a library like this is, it's should be considered low level and should try to support as much as it can.  I don't want to have to strand people if I don't have to, especially if it's not _that_ much work on my part. 

However,  I'd definitely like @tamizhvendan's opinion on forcing net472 at a minimum for this library.

NinoFloris:
Any remaining work for me to take care of?

TheAngryByrd:
> Any remaining work for me to take care of?

I think it looks good on your end, I just would like @tamizhvendan's opinion on dropping net461 since I'm still somewhat hesitant.

TheAngryByrd:
@NinoFloris I just fixed Build Checks in https://github.com/demystifyfp/FsToolkit.ErrorHandling/pull/98, would you might giving this a rebase? Thanks 😄

NinoFloris:
Done!

Issue #99, opened 2020-10-08
How to unwrap an async


JordanMarr:
I just upgraded from Cvdm.ErrorHandling to this.
In Cvdm.ErrorHandling, I could unwrap an `Async<Result<User, string>>` by using `let!` operator without piping into any of the AsyncResult or Result module conversion functions.

```
// tryGetUser = int -> Async<Result<User, string>>
let! myResult = tryGetUser(1) // returns a Result<User, string>
```

Now this fails because it's trying to look at the result type. But in this case, I don't want to fully unwrap it to be just a User -- I actually want it to be a Result<User, string> so that I can do some custom logic for the error case.

Is there a way to do this? I'm looking through the AsyncResult module, but I don't see anything relevant.

TheAngryByrd:
Yeah this is a known problem with how the binding overloads work in F#. 😞  This is discussed in  https://github.com/demystifyfp/FsToolkit.ErrorHandling/issues/88

Short answer, use the `async` CE or rework so you don't need to do this.

TheAngryByrd:
For a more concrete example, you can an async CE inline and do something like this:

```fsharp
let doTheThing () = asyncResult {
    let! user = async {
        let! myResult = tryGetUser(1)
        match myResult with
        | Ok user -> user
        | Error msg -> //custom errorHandling
    }
}
```

JordanMarr:
Shoot.
So then my options are to revert back to Cvdm version, or rework with async CE.
I suppose async CE is the better choice since Cvdm is no longer maintained.

TheAngryByrd:
Yeah I did this in my codebase in many places as well.  I should probably find the F# RFC for helping fix this issue so people 👍  it.

JordanMarr:
:.(

JordanMarr:
Would you consider adding this helper function to the AsyncResult module?

```
module AsyncResult =
    /// Awaits a result type without resolving it.
    let awaitResult (fn: Async<Result<_, _>>) =
        async {
            let! result = fn
            return Ok result
        }

// Usage
// tryGetUser = int -> Async<Result<User, string>>
let! myResult = tryGetUser(1) |> AsyncResult.awaitResult // returns a Result<User, string>
match myResult with
| Ok -> ...
| Error -> ...
```

TheAngryByrd:
I'm glad you found a solution that works for you but I don't think I would accept it.  Here are some of my reasons:

[...]

PR #100, opened 2020-10-08
Added AsyncResult.awaitResult helper function

JordanMarr:
See my issue #99


TheAngryByrd:
Sorry but like I said in the issue, I don't like this approach so I'm not going to accept this PR. I appreciate it a lot however.

Issue #106, opened 2020-11-16
Use affine tasks instead of non-affine


Swoorup:
Shouldn't the default behaviour be to use ply affine tasks instead of non-affine.

Reason being that current scheduler shouldn't be ignored. This allows having top-level control over degree of parallelism in which all child tasks would respect as well.

It's easy to override and use default scheduler even if using affine tasks than the other way around

TheAngryByrd:
No according to Microsoft general purpose libraries should be using ConfigureAwait(false) https://devblogs.microsoft.com/dotnet/configureawait-faq/

Swoorup:
Not sure I agree, that it should be the default. There is a minor performance without using `ConfigureAwait(false)` which I agree, but it means losing control of parallelism if subsequent child tasks are launched in parallel say in Asp.Net context.

It's possible to install the default/custom synchronization context when using the affine tasks, but not possible if using the non-affine version. 

In any case, I think it should be a conscious decision to choose the default. 
https://dev.to/noseratio/why-i-no-longer-use-configureawait-false-3pne

For myself since I am using this in asp.net the problem I am facing is inability to control parallelism for subsequent tasks launched as it freely uses all the threadpool resources, which is timing out other requests. 
https://blog.stephencleary.com/2017/03/aspnetcore-synchronization-context.html

TheAngryByrd:
Ok I read through those and they seem reasonable.  I don't have power currently so could you send a PR with this fix?

PR #107, opened 2020-11-17
Affine tasks

Swoorup:
Fixes https://github.com/demystifyfp/FsToolkit.ErrorHandling/issues/106

along with few other fixes

* Use affine tasks 
* Remove side by side github workflow hack


Swoorup:
@NinoFloris any reasons why methods like `task.Return` from ply return `Ply` type instead of tasks?

TheAngryByrd:
Ugh that side-by-side hack isn't working anymore 😢

TheAngryByrd:
> Ugh that side-by-side hack isn't working anymore 😢

Oh nice it was actually fixed and here's an example https://github.com/actions/setup-dotnet#usage

NinoFloris:
@Swoorup this is just the 'internal' exchange type, same goes for Taskbuilder (which has Step), it's the lowest common denominator between ValueTask and Task.

Swoorup:
Should fstoolkit be using ply as intermediate type at all, or always use task? I am finding it hard to work out compiler messages. The test are failing to build but the library builds.

Swoorup:
It appears fake doesn't allow changing the target to net50. Other than that this is ready for review. @TheAngryByrd

Swoorup:
Not sure how to fix the tooling issue
FSharp.Compiler.CompileOps+FileNameNotResolved' was thrown

Swoorup:
Reverted to just fix the linked issue

TheAngryByrd (approved):
Looks good to me. @NinoFloris wanna give it a once over as well?

src/FsToolkit.ErrorHandling.TaskResult/TaskOptionCE.fs
```diff
@@ -57,63 +56,66 @@ module TaskOptionCE =
...
         member inline _.Combine
           (computation1: Ply<Option<'T>>,
            computation2: unit -> Ply<Option<'U>>)
-          : Ply<Option<'U>> = uply {
-            match! computation1 with
-            | None -> return None
-            | Some _ -> return! computation2()
-          }
+          : Ply<Option<'U>> =
+            task {
```
NinoFloris, line 65: revert

src/FsToolkit.ErrorHandling.TaskResult/TaskOptionCE.fs
```diff
@@ -57,63 +56,66 @@ module TaskOptionCE =
...
-                match! computation() with
-                | Some _ as o ->
-                  result <- o
-                | None ->
-                  result <- None
-                  fin <- true
-              return result
-            }
+            : Ply<Option<'U>> =
+              task {
```
NinoFloris, line 92: revert

src/FsToolkit.ErrorHandling.TaskResult/TaskOptionCE.fs
```diff
@@ -57,63 +56,66 @@ module TaskOptionCE =
...
-                match! binder enumerator.Current with
-                | Some _ as o ->
-                  result <- o
-                | None ->
-                  result <- None
-                  fin <- true
-              return result
-            }
+            : Ply<Option<'U>> =
+              task {
```
NinoFloris, line 107: revert

src/FsToolkit.ErrorHandling.TaskResult/TaskResultCE.fs
```diff
@@ -30,66 +29,69 @@ module TaskResultCE =
...
     member inline _.Combine
         (computation1: Ply<Result<unit, 'TError>>,
          computation2: unit -> Ply<Result<'U, 'TError>>)
-        : Ply<Result<'U, 'TError>> = uply {
-          match! computation1 with
-          | Error e -> return Error e
-          | Ok _ -> return! computation2()
-        }
+        : Ply<Result<'U, 'TError>> =
+          task {
```
NinoFloris, line 43: revert

src/FsToolkit.ErrorHandling.TaskResult/TaskResultCE.fs
```diff
@@ -30,66 +29,69 @@ module TaskResultCE =
...
-          while not fin && guard() do
-            match! computation() with
-            | Ok x -> x
-            | Error _ as e ->
-              result <- e
-              fin <- true
-          return result
-        }
+        : Ply<Result<unit, 'TError>> =
+          task {
```
NinoFloris, line 70: revert

src/FsToolkit.ErrorHandling.TaskResult/TaskResultCE.fs
```diff
@@ -30,66 +29,69 @@ module TaskResultCE =
...
-          while not fin && enumerator.MoveNext() do
-            match! binder enumerator.Current with
-            | Ok x -> x
-            | Error _ as e ->
-              result <- e
-              fin <- true
-          return result
-        }
+        : Ply<Result<unit, 'TError>> =
+          task {
```
NinoFloris, line 84: revert

NinoFloris (changes requested):
The switch looks good. 

I'm not quite sure about all the renames from uply to task, as the return type that you get is not `Task` but `Ply`, it doesn't really aid readability this way but I'm not against it either.

One important change that should happen is every place where `uply { ... }` was swapped for `task { ... } |> task.ReturnFrom` should get reverted, this adds needless speedbump allocations going to Task + methodbuilders etc to then go straight back to Ply. 

The idea of uply {} is for it to be used as an internal type of the CE methods as much as possible, only to be mapped to the final type Task/ValueTask/Async in `Run`.

Either way, thanks for doing this :)

@TheAngryByrd I'm all for this change but there was a reason we didn't do this during the conversion to Ply in the first place. It should be very clear to users of the library currently depending on these CEs to automatically apply `ConfigureAwait(false)` that they can no longer expect it to do so. Meaning at every bind where they previously expected the CE to ignore any sync context or task scheduler it's now up to users to sprinkle `ConfigureAwait(false)` on those tasks before binding. If not then this release is bound to create deadlocks for users of gui or older aspnet frameworks.

TheAngryByrd:
> @TheAngryByrd I'm all for this change but there was a reason we didn't do this during the conversion to Ply in the first place. It should be very clear to users of the library currently depending on these CEs to automatically apply ConfigureAwait(false) that they can no longer expect it to do so. Meaning at every bind where they previously expected the CE to ignore any sync context or task scheduler it's now up to users to sprinkle ConfigureAwait(false) on those tasks before binding. If not then this release is bound to create deadlocks for users of gui or older aspnet frameworks. 

Yeah I'm sure I'm gonna get an issue about this problem too. Might have to make a separate package with unaffine vs affine but we'll see if people complain first before I go fixing problems that may not exist.  Additionally the older versions aren't going anywhere for now.

TheAngryByrd (changes requested):
Please address Nino's comments.

src/FsToolkit.ErrorHandling.TaskResult/TaskOptionCE.fs
Swoorup: @NinoFloris I was under the impression the unsafe version ignores the context as well? Or are they just restricted to all safe or unsafe coming from non-affine modules?
https://github.com/crowded/ply#execution-bubble

NinoFloris: The execution context is not the same as the sync context/scheduler. The unsafe CEs don't capture the previous execution context at the start of the CE (captured to be put back when the CE sets a result on the Task), which can cause AsyncLocals to leak into the outer scope, that's the gist of safe vs unsafe. However when Run uses a safe builder to execute the internal ply returning continuation the execution context will be properly captured and reset by that safe builder.

For the sync context/scheduler all that matters is Affine/NonAffine.

Swoorup: Thanks for clearing that up

Swoorup:
@NinoFloris @TheAngryByrd Have addressed the feedback

TheAngryByrd:
Thank you!

TheAngryByrd (approved):

TheAngryByrd:
@Swoorup Take `2.0.0-beta002` for a spin

Swoorup:
Pre 2.0.0-beta002, the following code would give 

```fsharp
#r "nuget: Ply"
#r "nuget: FsToolkit.ErrorHandling.TaskResult,2.0.0-beta002"

open System
open System.Threading
open System.Threading.Tasks
open FSharp.Control.Tasks
open FsToolkit.ErrorHandling
open FsToolkit.ErrorHandling

let a () =
  task {
    printfn "From taskCE Scheduler Id: %A" TaskScheduler.Current.Id
    let! a = 
      taskResult { 
        do! Task.Delay(TimeSpan.FromSeconds(1.0)) 
        printfn "From taskResultCE Scheduler Id: %A" TaskScheduler.Current.Id
      }
    return ()
  } 

let Scheduler = ConcurrentExclusiveSchedulerPair(TaskScheduler.Default, 1, 1)
printfn "Main Scheduler ID: %A" TaskScheduler.Current.Id
printfn "Scheduler ID: %A" Scheduler.ExclusiveScheduler.Id

Task.Factory.StartNew(a, CancellationToken.None, TaskCreationOptions.None, Scheduler.ExclusiveScheduler) |> ignore
Console.ReadLine()
```

#### Pre 2.0.0-beta002
image

#### 2.0.0-beta002
image

Works well now.

PR #120, opened 2021-03-11
Fix: Build Issues

no1melman:
attempting to fix actions build issues


no1melman:
@TheAngryByrd I noticed a build issue with my last commit - I think this is the right fix - but will leave that for you to decide.

TheAngryByrd:
I think this will work _for now_ but we should align the version number in the global.json and build.yml https://github.com/demystifyfp/FsToolkit.ErrorHandling/blob/443dd1016ca7fd98cc52ca99c08932c6a4d10896/.github/workflows/build.yml#L25 or else this will happen again when Microsoft released a new build

PR #129, opened 2021-05-14
Update fable 3

TheAngryByrd:


package.json
```diff
@@ -1,11 +1,13 @@
 {
+  "type": "module",
   "devDependencies": {
+    "esm": "^3.2.25",
     "fable-compiler": "^2.9.0",
     "fable-splitter": "^2.1.13",
```
alfonsogarciacaro, line 6: Both fable-compiler and fable-splitter shouldn't be need now

package.json
```diff
@@ -1,11 +1,13 @@
...
-    "mocha": "^6.1.4"
+    "mocha": "^8.3.2"
   },
   "scripts": {
-    "pretest": "fable-splitter tests/FsToolkit.ErrorHandling.Tests -o dist/tests --commonjs",
+    "pretest": "dotnet fable tests/FsToolkit.ErrorHandling.Tests -o dist/tests",
     "test": "mocha dist/tests",
     "watch-build": "fable-splitter  tests/FsToolkit.ErrorHandling.Tests -o dist/tests --commonjs -w",
```
alfonsogarciacaro, line 12: "watch-build" can also be updated as "dotnet fable watch tests/FsToolkit.ErrorHandling.Tests -o dist/tests"

Also, if you add a `"postinstall": "dotnet tool restore"` script, it will be possible to run the js tests on a fresh clone just by typing "npm i && npm test".

alfonsogarciacaro (commented):
Would it be possible to also add `"rollForward": "minor"` to global.json so we can use a newer dotnet sdk version within the 5.x range?

PR #132, opened 2021-05-26
Added Async, Task and Job overloads for the relavant Option CEs to resolve #117

### LINKED ISSUE #117: Enable asyncOption CE to bind Async<'a>
Currently it can be annoying to use `asyncOption` because I have to append `|> Async.map Some` everywhere I bind a computation that does not return `Async<'a option>` but just `Async<'a>`.

It would be great if the `Bind` member was overloaded to accept a plain `Async<'a>`. (This must be an extension member to avoid overload resolution problems.)

Jmaharman:
These commits resolve the issue #117. @baronfel guided me towards Source after previously working with @TheAngryByrd to simplify the TaskResult CE. I really liked his description of what Source does, so I'm placing it here before it's lost to the black hole of Slack:

Calls to Source are applied automatically by the compiler at points where values can enter the bind chain, and that leaves less work for you as the CE implementor because then you just have to implement Bind/Return/ReturnFrom in terms of your 'core' data type (in your case Async<Option<'t>>) and let the Source overloads handle the unification of the input data types into your 'core' data type.

### DIFF
--- a/src/FsToolkit.ErrorHandling.JobResult/JobOptionCE.fs
+++ b/src/FsToolkit.ErrorHandling.JobResult/JobOptionCE.fs
@@ -13,29 +13,9 @@ module JobOptionCE =
       job.Return <| option.Return value
 
     member __.ReturnFrom
-        (asyncResult: Async<Option<_>>)
-        : Job<Option<_>> =
-      asyncResult |> Job.fromAsync
-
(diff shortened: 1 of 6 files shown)


TheAngryByrd (approved):
LGTM. Thanks!!!

PR #135, opened 2021-08-01
Extra Task, ValueTask, Ply CE sources

kerams:
Also adds `MergeSources` for `OptionResult`.

### DIFF
--- a/src/FsToolkit.ErrorHandling.TaskResult/Task.fs
+++ b/src/FsToolkit.ErrorHandling.TaskResult/Task.fs
@@ -12,11 +12,18 @@ module Task =
       return! f x
   }
 
+  let bindV (f : 'a -> Task<'b>) (x : ValueTask<'a>) = task {
+      let! x = x
+      return! f x
+  }
+
   let apply f x =
     bind (fun f' ->
       bind (fun x' -> singleton(f' x')) x) f
 
   let map f x = x |> bind (f >> singleton)
+
+  let mapV f x = x |> bindV (f >> singleton)
 
   let map2 f x y =
     (apply (apply (singleton f) x) y)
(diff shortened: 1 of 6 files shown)


src/FsToolkit.ErrorHandling.TaskResult/Task.fs
```diff
@@ -12,12 +12,19 @@ module Task =
...
+  }
+
   let apply f x =
     bind (fun f' ->
       bind (fun x' -> singleton(f' x')) x) f
 
   let map f x = x |> bind (f >> singleton)
 
+  let mapV f x = x |> bindV (f >> singleton)
+
```
kerams, line 27: Not sure about the general utility of these 2 functions. If you prefer, I could remove them and just do

```fsharp
task {
    let! t = t
    return Ok t // and Some t
}
```

in the `Source` methods for `ValueTask<'t>`. They'd also run a tiny bit faster as a result.

TheAngryByrd: > They'd also run a tiny bit faster as a result.

Yeah I've been wondering if doing pure functions is better or not. If you could show this via benchmarks.net I'd highly appreciate it! 😄 


I'm going to accept this PR as the slight speedup isn't a blocker for getting this awesome functionality out the door.

TheAngryByrd (changes requested):
CI failure:

```
Error: /home/runner/work/FsToolkit.ErrorHandling/FsToolkit.ErrorHandling/src/FsToolkit.ErrorHandling.TaskResult/TaskOptionCE.fs(147,21): error FS1113: The value 'Source' was marked inline but its implementation makes use of an internal or private function which is not sufficiently accessible [/home/runner/work/FsToolkit.ErrorHandling/FsToolkit.ErrorHandling/src/FsToolkit.ErrorHandling.TaskResult/FsToolkit.ErrorHandling.TaskResult.fsproj]
```

PR #140, opened 2021-10-19
Canonicalize option type annotations

cmeeren:
This changes all `Option<_>` type annotations to the [idiomatic `_ option` form](https://docs.microsoft.com/en-us/dotnet/fsharp/style-guide/formatting#for-types-prefer-prefix-syntax-for-generics-foot-with-some-specific-exceptions).

I have long been annoyed at `Option<_>` Intellisense type hints across my solution when I use `asyncOption`. Suddenly realized I could probably do something about it fairly easily by creating a trivial PR. 😅


TheAngryByrd:
Ok fixed CI I think : https://github.com/demystifyfp/FsToolkit.ErrorHandling/pull/141

PR #146, opened 2021-10-28
issue-144-sequence-result-m

njlr:
* Adds Seq module for lazy result traversal


njlr:
Should work after merging https://github.com/demystifyfp/FsToolkit.ErrorHandling/pull/149

TheAngryByrd (approved):
Thanks for all the help!

PR #157, opened 2021-11-27
Update build.sh

njlr:
I think this will fix the CI build. 

My understanding is that `.paket` should be in `.gitignore` and it needs to be generated by `paket restore`.


TheAngryByrd:
Thanks!!!

PR #169, opened 2022-03-29
Native Tasks for TaskResult, TaskOption, and TaskResultOption.

### LINKED ISSUE #142: Native tasks
Now that [F# 6.0 has native task {} support](https://devblogs.microsoft.com/dotnet/whats-new-in-fsharp-6/#making-f-faster-and-more-interopable-with-task) we should figure out how/when to integrate this.  

- I'm assuming this would force this library to make consumers go to F# 6.0.  
- Additionally, one of the reasons `TaskResult` library is separate is because it relied on upstream dependency and I always try to minimize the dependency tree that consumers need to use a library.  The other is it makes it easier to deal with Fable instead of having large portions ifdef'd out.  The first reason would no longer matter but the second still does unless Fable implements some shim for `task {}`.

TheAngryByrd:
## Proposed Changes

Closes #142 

## Types of changes

_Put an `x` in the boxes that apply_

## Checklist

## Further comments

To make all the existing tests to pass. I had to implement my own versions of `While` and `Combine` as [described by BDFN](https://github.com/dotnet/fsharp/discussions/12792#discussioncomment-2463397).  

Testing the Dynamic path isn't easy currently.  Commenting out the state machine part of `Run` and delegating directly to `RunDynamic` is how I tested some implementations.  [The only other way I've seen](https://github.com/dotnet/fsharp/blob/f45ef819b5310de521645dbc68ccbfc57f8edda4/tests/FSharp.Core.UnitTests/FSharp.Core/Microsoft.FSharp.Control/TasksDynamic.fs) is implement additional CE's and force it to not inline the state machine. 

~~I implemented `backgroundTaskResult` and `backgroundTaskOption` but I haven't added tests around them.~~ Tests have been added and I've implemented a corresponding version using Ply in the netstandard2.0 TFM.

~~`taskResultOption` is implemented in the laziest way possible by re-using the native Task resumption code.~~ This now uses it's own resumable code implemenation.


TheAngryByrd:
While I haven't figured out the best way yet to test the Dynamic path, i'm going to merge this to get feedback from people in the field.

TheAngryByrd:
I address some of this here: 

https://github.com/demystifyfp/FsToolkit.ErrorHandling/issues/175#issuecomment-1099425275

additionally, we still do have Ply as a dependency for older .net versions still around

PR #174, opened 2022-04-04
Add explicit type parameters to ignore functions

### LINKED ISSUE #173: Add explicit type parameters to ignore functions
I'd like for `ignore` functions like `AsyncResult.ignore` (and perhaps `ignoreError`?) to accept explicit type parameters, just like F#'s `ignore` function.

My reasoning is that ignoring input is a "dangerous" operation under refactoring. For example:

```f#
doSomething |> ignore<int>
```

If you change `doSomething` so that it no longer returns `int`, then that may indicate that the code that previously ignored its result should now do something else with it. If you just specify `ignore`, you have to remember to check this yourself. If you specify an explicit type parameter, the compiler checks this for you and you have to decide in each case whether to ignore the new return type or do something else.

It is currently not possible to use explicit type parameters with e.g. `AsyncResult.ignore` because the definition does not use explicit type parameters. If this is changed, we can use explicit type parameters, e.g. `AsyncResult.ignore<int, _>`.

cmeeren:
Fixes #173

If anything needs fixing, feel free to make any edits you want, or merge and fix up later, or anything, really. 🙂

### DIFF
--- a/src/FsToolkit.ErrorHandling.JobResult/JobResult.fs
+++ b/src/FsToolkit.ErrorHandling.JobResult/JobResult.fs
@@ -97,7 +97,7 @@ module JobResult =
         result |> Job.bind (Result.either ok ifErrorFunc)
 
     /// Replaces the wrapped value with unit
-    let inline ignore jr = jr |> map ignore
+    let inline ignore<'ok, 'error> (jr: Job<Result<'ok, 'error>>) = jr |> map ignore<'ok>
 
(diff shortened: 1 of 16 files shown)


TheAngryByrd (changes requested):
Looks great!

I think this just needs a `./build.sh FormatCode` and it should be all good!

cmeeren:
Done!

PR #186, opened 2022-10-19
Add helper functions for options wrapped in a TaskResult, AsyncResult, and JobResult

### LINKED ISSUE #184: Task wrapped result + option helpers
**Is your feature request related to a problem? Please describe.**
Helper functions to work with ``Task<Result<'success option, 'error>>``

Example:
```fsharp
type Error =
  | UsernameTaken
  | DatabaseError of DbError

taskResult {
  do!
    // getUserByUsername : string -> Task<Result<User option, DbError>>
    getUserByUsername createAccountRequest.Username // 
    |> TaskResult.mapError DatabaseError
    |> TaskResult.bindRequireNone UsernameTaken
}
```

**Describe the solution you'd like**
Helper functions that combine `bind` + `requireSome`, `requireNone`, etc... 

```fsharp
// name needs work :P
let inline bindRequireSome (error: 'error) (taskResultOption: Task<Result<'success option, 'error>>) =
    taskResultOption
    |> bind (Result.requireSome error >> Task.singleton)
```

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

I would like to create a PR for this.

sheridanchris:
## Proposed Changes

Add util functions to work with `Task<Result<'success option, 'error>>` values.
This will close #184

## Types of changes

## Checklist

- [ ] I have added the necessary documentation (if appropriate)

I'm unsure about the last one.

## Further comments

I wonder if there are similar enhancement opportunities in other modules. Not only working with options wrapped in task results but also options wrapped in normal results. Also, potentially applying this to additional functions (other than `requireNone` and `requireSome`). I also wonder if a new module named `TaskResultOption` should be created to complement the existing `TaskResult` and `TaskOption` modules.

### DIFF
--- a/tests/FsToolkit.ErrorHandling.TaskResult.Tests/TaskResult.fs
+++ b/tests/FsToolkit.ErrorHandling.TaskResult.Tests/TaskResult.fs
@@ -827,15 +827,21 @@ let TaskResultOperatorTests =
 [<Tests>]
 let TaskResultBindRequireTests =
     testList "TeskResult Bind + Require Tests" [
-        testCase "bindRequireNone"
-        <| fun _ ->
-            let user = Some "john_doe" |> Ok |> Task.singleton
-            let error = user |> TaskResult.bindRequireNone "User exists"
-            error |> Expect.hasTaskErrorValueSync "User exists"
+        testCaseTask "bindRequireNone"
+        <| fun _ -> task {
+            do!
+                Some "john_doe"
+                |> TaskResult.ok
+                |> TaskResult.bindRequireNone "User exists"
+                |> Expect.hasTaskErrorValue "User exists"
+           }
 
-        testCase "bindRequireSome"
-        <| fun _ ->
-            let user = Some "john_doe" |> Ok |> Task.singleton
-            let error = user |> TaskResult.bindRequireSome "User doesn't exist"
-            error |> Expect.hasTaskOkValueSync "john_doe"
+        testCaseTask "bindRequireSome"
+        <| fun _ -> task {
+            do!
+                Some "john_doe"
+                |> TaskResult.ok
+                |> TaskResult.bindRequireSome "User doesn't exist"
+                |> Expect.hasTaskOkValue "john_doe"
+           }
     ]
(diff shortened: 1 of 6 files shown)


TheAngryByrd:
> I wonder if there are similar enhancement opportunities in other modules. Not only working with options wrapped in task results but also options wrapped in normal results. Also, potentially applying this to additional functions (other than requireNone and requireSome).

Yeah I would say add these similar things to the [AsyncResult](https://github.com/demystifyfp/FsToolkit.ErrorHandling/blob/master/src/FsToolkit.ErrorHandling/AsyncResult.fs) and [JobResult](https://github.com/demystifyfp/FsToolkit.ErrorHandling/blob/master/src/FsToolkit.ErrorHandling.JobResult/JobResult.fs) modules.

While there is quite a bit of `require` type functions I think it's fine on this pass to just do the `requireNone` `requireSome` cases.

>  I also wonder if a new module named TaskResultOption should be created to complement the existing TaskResult and TaskOption modules.

We do have [this module](https://github.com/demystifyfp/FsToolkit.ErrorHandling/blob/master/src/FsToolkit.ErrorHandling.TaskResult/TaskResultOption.fs) already but it's semantics try to keep the happy path of `Ok(Some foo)` and not folding into just `Ok(foo)` so I'm unsure how it would fit within that modules concepts.

tests/FsToolkit.ErrorHandling.TaskResult.Tests/TaskResult.fs
```diff
@@ -823,3 +823,19 @@ let TaskResultOperatorTests =
             )
             |> Expect.hasTaskOkValueSync (PostId newPostId)
     ]
+
+[<Tests>]
+let TaskResultBindRequireTests =
+    testList "TeskResult Bind + Require Tests" [
+        testCase "bindRequireNone"
```
TheAngryByrd, line 830: Should make these `testCaseTask` as the expect statement you'd want to `return!`/`do!` to ensure it actually runs the expect code rather than fire & forget.

TheAngryByrd (changes requested):

tests/FsToolkit.ErrorHandling.TaskResult.Tests/TaskResult.fs
sheridanchris: Fixed, thank you!

TheAngryByrd (approved):
LGTM

Issue #191, opened 2022-11-07
Should List.sequenceValidationA preserve order?


ursenzler:
**Describe the bug**
I expected this test to succeed, but it failed because the order of the errors was reversed.

[<Fact>]
let ``List.sequenceValidation preserves order of errors`` () =
    let result = [ Result.Error [ "a" ]; Result.Error [ "b" ] ] |> List.sequenceValidationA
    test <@ result = Result.Error [ "a"; "b" ] @>

**To Reproduce**
Run above test

**Expected behavior**
Order should be preserved

**Additional context**
It's absolutely okay when you say that this behaviour is by design or because of performance reasons. Then I'll reverse the errors in my code. I just want to ensure that it's not a "bug" that gets fixed eventually, and my errors are wrong again.

TheAngryByrd:
Definitely feels like a missed `List.rev`.  Happy to take a contribution for fixing this!

ursenzler:
Ok, I'll work on it.

PR #192, opened 2022-11-08
List.traverseValidationA and List.sequenceValidationA now preserve the order of errors

ursenzler:
## Proposed Changes

Preserve the order of errors when executing List.traverseValidationA - and as a consequence for List.sequenceValidationA.
Fixes [Should List.sequenceValidationA preserve order? #191](https://github.com/demystifyfp/FsToolkit.ErrorHandling/issues/191)

## Types of changes
- [x] Bugfix (non-breaking change which fixes an issue)

Could however be a breaking change because it changes the behaviour!

## Checklist

## Further comments

I think this is rather straight forward :-) 

Let me know if something needs to be changed. E.g. I didn't format it with Fantomas because other parts of the file would have been changed as well.

### DIFF
--- a/src/FsToolkit.ErrorHandling/List.fs
+++ b/src/FsToolkit.ErrorHandling/List.fs
@@ -106,7 +106,12 @@ module List =
             | Ok _, Error errs
             | Error errs, Ok _ -> traverseValidationA' (Error errs) f xs
 
-    let traverseValidationA f xs = traverseValidationA' (Ok []) f xs
+    let traverseValidationA f xs =
+        traverseValidationA' (Ok []) f xs
+        |> fun r ->
+            match r with
+            | Ok _ -> r
+            | Error e -> e |> List.rev |> Result.Error
 
     let sequenceValidationA xs = traverseValidationA id xs
 
--- a/src/FsToolkit.ErrorHandling/List.fs
+++ b/src/FsToolkit.ErrorHandling/List.fs
@@ -89,7 +89,7 @@ module List =
         match xs with
         | [] ->
             state
-            |> Result.map List.rev
+            |> Result.eitherMap List.rev List.rev
         | x :: xs ->
             let fR = f x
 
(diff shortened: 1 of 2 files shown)


src/FsToolkit.ErrorHandling/List.fs
```diff
@@ -106,7 +106,12 @@ module List =
             | Ok _, Error errs
             | Error errs, Ok _ -> traverseValidationA' (Error errs) f xs
 
-    let traverseValidationA f xs = traverseValidationA' (Ok []) f xs
+    let traverseValidationA f xs =
```
TheAngryByrd, line 109: I think there's a similar fix for this type of behavior where you can keep it in `traverseValidationA'`

https://github.com/demystifyfp/FsToolkit.ErrorHandling/blob/31784fa747d7a7dfd5a03c61f8a3ef436c117d51/src/FsToolkit.ErrorHandling/List.fs#L57

TheAngryByrd (changes requested):
Minor Change but looks good otherwise.

TheAngryByrd:
> I didn't format it with Fantomas because other parts of the file would have been changed as well.

That's fine. I'm used to it by now

src/FsToolkit.ErrorHandling/List.fs
ursenzler: Yes, that is simpler.

ursenzler:
I formatted List.fs with Fantomas, but it's not in an accepted format. I'm a bit lost here.

TheAngryByrd:
hmm did you run `build.cmd/sh formatcode`  or did you run it from your editor? I don't know if VS or Rider pickup the dotnet tool correctly, which the FAKE build is using the dotnet-tool.

TheAngryByrd:
Thank you!

ursenzler:
Thanks for this excellent library!!

PR #213, opened 2023-03-27
Update fantomas

nojaf:
v6 is safe to use and I would like to dogfood it a bit.
I'm not sure what your stance is on computation expressions.
Let me know if I need to turn on [fsharp_newline_before_multiline_computation_expression](https://fsprojects.github.io/fantomas/docs/end-users/Configuration.html#fsharp_newline_before_multiline_computation_expression).


TheAngryByrd:
Thanks!

PR #219, opened 2023-07-09
Bind 'Mappings' Documentation

1eyewonder:
## Proposed Changes

I felt like adding a section on what types can implicity bind to each other would be useful. I didn't know you could do this immediately when I started using the library, so I felt it may make this feature more apparent to new users. It may also prove useful as reference material when users are trying to combine CEs in their codebases.

## Types of changes

## Checklist

- [ ] Build and tests pass locally
- [X] I have added necessary documentation (if appropriate)


TheAngryByrd (approved):
You rock 🎸

PR #222, opened 2023-07-11
Documentation Updates

1eyewonder:
## Proposed Changes

When doing through documentation, I noticed there were some items which were listed, but not applicable to certain modules as well as unlisted functions in modules. I created a 'PR' markdown document which will be shown on the documentation site. The reasoning behind this is to allow users to know the feature is supported, but we just haven't gotten to the documentation yet.

Additionally, I started adding examples to the Result module. I feel the documentation is lacking in examples overall so I figured brining over some of our unit tests helps provide visuals for users who may not grasp the function signatures at first glance.

## Checklist
- [X] I have added necessary documentation (if appropriate)


TheAngryByrd:
This is so amazing, thank you!

Issue #225, opened 2023-08-04
Support ValueOption in FsToolkit.Errorhandling.asyncOption


Xyncgas:
currently we can use ```let! x = ``` with say ``` int option``` but not ```int voption```
it seems voptions can reduce memory allocation with option types

TheAngryByrd:
Yep I agree this is a good idea. Would you like to submit a pull request for it?

Xyncgas:
Why do I have to do everything

Xyncgas:
Maybe instead of making it possible by doing what OptionCE did, I can make asyncOption CE to use voption instead of converting voption to option

Xyncgas:
in the future

PR #226, opened 2023-08-04
Add Async<'value voption> support to asyncOptionCE

Xyncgas:
We can currently use ```let!``` to bind to either option or value option in ```option {}``` computation expression, I am making asyncOption to be able to do this too, it couldn't before I made it to.

it does so by converting your voption to option, which is how OptionCE does it
https://github.com/demystifyfp/FsToolkit.ErrorHandling/issues/225#issuecomment-1665588509

### DIFF
--- a/tests/FsToolkit.ErrorHandling.Tests/AsyncOptionCE.fs
+++ b/tests/FsToolkit.ErrorHandling.Tests/AsyncOptionCE.fs
@@ -105,7 +105,7 @@ let ``AsyncOptionCE bind Tests`` =
 
         }
 
-        testCaseAsync "Bind ValueSome Result"
+        testCaseAsync "Bind ValueSome AsyncOption"
         <| async {
             let innerData = "Foo"
             let data = ValueSome innerData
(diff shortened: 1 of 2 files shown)


TheAngryByrd (changes requested):
Thanks for this! Could you also add corresponding tests in the test suite? Tests for let! and return! should be sufficient.

Xyncgas:
Added tests

tests/FsToolkit.ErrorHandling.Tests/AsyncOptionCE.fs
```diff
@@ -35,6 +35,15 @@ let ``AsyncOptionCE return! Tests`` =
             Expect.equal actual (data) "Should be ok"
         }
 
+        testCaseAsync "Return ValueSome Result"
```
TheAngryByrd, line 38:
```suggestion
        testCaseAsync "Return ValueSome AsyncOption"
```

tests/FsToolkit.ErrorHandling.Tests/AsyncOptionCE.fs
```diff
@@ -96,6 +105,21 @@ let ``AsyncOptionCE bind Tests`` =
 
         }
 
+        testCaseAsync "Bind ValueSome Result"
```
TheAngryByrd, line 108:
```suggestion
        testCaseAsync "Bind ValueSome AsyncOption"
```

TheAngryByrd (changes requested):

PR #231, opened 2023-10-16
Added traverse and sequence functions for Option

1eyewonder:
## Proposed Changes

Added traverse and sequence functions for both Option and AsyncOption types. I made this due to discussion in the discord channel and thought it may be pretty straight forward.

Functions added:

- List.traverseOptionM
- List.traverseAsyncOptionM
- List.sequenceOptionM
- List.sequenceAsyncOptionM

## Types of changes

## Checklist

- [X] I have added tests that prove my fix is effective or that my feature works (if appropriate)

## Further comments

If the behavior matches what is expected, let me know and I can create the documentation as well 😄 

Do we think the applicative versions should be added as well? I didn't see an immediate use case but I know `pblasucci` in the discord was the one with the use case

### DIFF
--- a/src/FsToolkit.ErrorHandling/List.fs
+++ b/src/FsToolkit.ErrorHandling/List.fs
@@ -159,6 +159,7 @@ module List =
 
     let sequenceOptionM xs = traverseOptionM id xs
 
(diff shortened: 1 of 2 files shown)


TheAngryByrd (approved):
Thank you!!

PR #242, opened 2023-12-09
Add TaskResult.foldResult

### LINKED ISSUE #218: Can we reinstate `TaskResult.foldResult` and the like?
Deleting publicly surfaced functions has the big downside that people cannot update their libraries anymore. In this commit (https://github.com/demystifyfp/FsToolkit.ErrorHandling/commit/ac33c8588887319198e02624968ca17faa2b802a) most (but not all, `AsyncResult.foldResult` still exists) `foldResult`s were removed, apparently because F# Core introduced them? This may be true for `Result.fold`, but certainly not for the compound types.

Instead, if F# introduced a version of some functionality, if you have a different name, I propose to just alias the F# Core function (if it is already available in the lowest supported F# Core, otherwise just copy its implementation, which I think you've done in the past).

In the case of `TaskResult.foldResult` , `CancellableTask.foldResult` and `JobResult.foldResult`, they just disappeared from FsToolkit, and `Result.fold` as well. It led (without me knowing) to one team not updating the library anymore as it broke the code (and it was non-trivial to them how to resolve).

To be fair, you kept `Result.defaultWith`, which existed for similar arguments (see #195, #199).

### To reproduce

Upgrade from an older version to a newer and code starts failing to compile.

### Expected behavior

More conservative backward compat behavior, if possible (and you know me, I'm willing to help with this! Just ping me)

PS: the documentation still shows that these functions exist: https://demystifyfp.gitbook.io/fstoolkit-errorhandling/fstoolkit.errorhandling.taskresult/taskresult/foldresult

Maxumka:
## Proposed Changes

resolve #218 

## Types of changes

## Checklist

### DIFF
--- a/src/FsToolkit.ErrorHandling.TaskResult/TaskResult.fs
+++ b/src/FsToolkit.ErrorHandling.TaskResult/TaskResult.fs
@@ -250,3 +250,10 @@ module TaskResult =
             Result.requireNone error
             >> Task.singleton
         )
+
+    let inline foldResult
+        ([<InlineIfLambda>] onSuccess: 'input -> 'output)
+        ([<InlineIfLambda>] onError: 'inputError -> 'output)
+        (input: Task<Result<'input, 'inputError>>)
+        : Task<'output> =
+        Task.map (Result.either onSuccess onError) input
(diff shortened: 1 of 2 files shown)


TheAngryByrd (approved):
Thank you!

Note: I realized we weren't running tests in CI so I had to rebase your branch after I fixed it and force push.

PR #243, opened 2023-12-09
Add support voption to require function

Maxumka:
## Proposed Changes

@TheAngryByrd Hi.

In my company we use voption and to convert to result we need to write our own functions for different projects. For convenience I would like to have them in a package at once

## Types of changes

## Checklist

## Further comments

Also added bind functions to require voption


TheAngryByrd (approved):
Thank you! And all the documentation is great :)

Note: I realized we weren't running tests in CI so I had to rebase your branch after I fixed it and force push.

Issue #254, opened 2024-02-14
feat(Seq): Add sequenceResultA


bartelink:
There's a List.sequenceResultA

I'd find use for a `Seq.sequenceResultA: Result<'t, 'e> seq -> Result<'t[], 'e[]>`

This would align with how Async.Parallel works (accumulate outputs in a ResizeArray, yield an array) for efficiency

Example desired consumption

```
let res: Result<EmailAddress[], string> =
    seq { for i in Seq.distinct inputAddrs -> EmailAddress.tryParse i |> Result.requireSome i }
    |> Seq.sequenceResultA
    |> Result.mapError (fun es -> $"""Invalid inputs: {es |> String.concat ","}""" )
```

Example current workaround:

```
let res: Result<EmailAddress list, string> =
    [ for i in Seq.distinct inputAddrs -> EmailAddress.tryParse i |> Result.requireSome i ]
    |> List.sequenceResultA
    |> Result.mapError (fun es -> $"""Invalid inputs: {es |> String.concat ","}""" )
```

(This is literally what I have, and I have limited awareness of the Validation modules etc so maybe they offer a better workaround, but it does seem that [in general there's a hole in the completeness of the operations](https://stackoverflow.com/questions/50789065/turn-list-of-result-into-result-of-list-inside-a-computation-expression#comment137501687_71081606) hence raising this)

TheAngryByrd:
Yep I agree this is a good idea. Would you like to submit a pull request for it?

TheAngryByrd:
> I find the semantics of the laziness of that hard to reason abou

Yeah it seems like it's it's a Seq.fold, so I don't think it's going to be something that can lazily be evaluated. 

> Seq one could be obsoleted

Not necessarily because there are other enumerables beyond F#'s Array/List in the BCL and it's handy for those.

bartelink:
Updates from spiking:
1. Fantomas default settings drive me wild (one pipe forcing a newline and other such nonsense). It's a real turnoff (and paket is IMO a bad idea for libraries). The ones TaskSeq are pretty tolerable. [The ones in FSharp.AWS.DyamoDB, which are a fork with 3 or 4 tweaks from TaskSeq, are very good](https://github.com/fsprojects/FSharp.AWS.DynamoDB/pull/73)
[...]

PR #255, opened 2024-02-15
feat(Seq)!: Add sequenceResultA, align sequenceResultM

bartelink:
Implements #254 
- Add `Seq.sequenceResultA`
- align signature of `sequenceResultM` with that of `sequenceResultA` (yield an Array, rather than accumulating in a `list` and upcasting via `List.toSeq`
- add docs for `sequenceResultM`
- remove FSharp.xml content file (noise when searching in Rider)

### DIFF
--- a/build/build.fsproj
+++ b/build/build.fsproj
@@ -4,6 +4,8 @@
     <OutputType>Exe</OutputType>
     <TargetFramework>net7.0</TargetFramework>
     <IsPackable>false</IsPackable>
+    <!-- <NoWarn>NU1904</NoWarn>-->
+    <TreatWarningsAsErrors>false</TreatWarningsAsErrors>
   </PropertyGroup>
   <ItemGroup>
     <Compile Include="DotEnv.fs" />
--- a/src/FsToolkit.ErrorHandling/Seq.fs
+++ b/src/FsToolkit.ErrorHandling/Seq.fs
@@ -19,3 +19,19 @@ let sequenceResultM (xs: seq<Result<'t, 'e>>) : Result<'t[], 'e> =
             err <- e
 
     if ok then Ok(acc.ToArray()) else Error err
+
+let sequenceResultA (xs: seq<Result<'t, 'e>>) : Result<'t[], 'e[]> =
+    if isNull xs then
+        nullArg (nameof xs)
+
+    let oks = ResizeArray()
+    let errs = ResizeArray()
+
+    for x in xs do
+        match x with
+        | Ok r -> oks.Add r
+        | Error e -> errs.Add e
+
+    match errs.ToArray() with
+    | [||] -> Ok(oks.ToArray())
+    | errs -> Error errs
--- a/src/FsToolkit.ErrorHandling/Seq.fs
+++ b/src/FsToolkit.ErrorHandling/Seq.fs
@@ -1,19 +1,21 @@
-namespace FsToolkit.ErrorHandling
-
 [<RequireQualifiedAccess>]
-module Seq =
+module FsToolkit.ErrorHandling.Seq
 
-    let sequenceResultM (xs: seq<Result<'t, 'e>>) : Result<'t seq, 'e> =
-        let rec loop xs ts =
-            match Seq.tryHead xs with
-            | Some x ->
-                x
-                |> Result.bind (fun t -> loop (Seq.tail xs) (t :: ts))
-            | None ->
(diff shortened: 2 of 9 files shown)


bartelink:
Open to removing anything you consider an overstep - I'm not enjoying the fantomas ruleset so there is some letter of the law interpretations.

Please push back on anything that's not right!

paket.dependencies
```diff
@@ -65,7 +65,8 @@ nuget BenchmarkDotNet.Diagnostics.Windows 0.13.1
 group Build
     source https://api.nuget.org/v3/index.json
     storage: none
-    nuget FSharp.Core
+
+    nuget FSharp.Core ~> 7
```
bartelink, line 69: this is the material change - v7 and later don't have erroneous xmldoc as content in the package
can alternately override with a `content: none`

paket.dependencies
```diff
@@ -1,46 +1,44 @@
...
-nuget Ply
-nuget Hopac
-nuget FSharp.Control.AsyncSeq
-nuget Fable.Core >= 4.2.0
 framework: netstandard2.0, net6.0
 storage: none
 condition: netstandard2_0
 
-
+nuget FSharp.Core >= 4.7.2 content: none
```
bartelink, line 15: drop erroneous content files

src/FsToolkit.ErrorHandling/Seq.fs
```diff
@@ -1,19 +1,37 @@
...
-                Ok(
-                    List.rev ts
-                    |> List.toSeq
-                )
-
-        // Seq.cache prevents double evaluation in Seq.tail
-        loop (Seq.cache xs) []
+module FsToolkit.ErrorHandling.Seq
+
+let sequenceResultM (xs: seq<Result<'t, 'e>>) : Result<'t[], 'e> =
```
bartelink, line 4: note we return an array for consistency with sequenceResultA
if people want arrays or allocations, List.sequenceResultM covers that

TheAngryByrd: I'd prefer not to make breaking changes here unless you want to retarget for [v5](https://github.com/demystifyfp/FsToolkit.ErrorHandling/pull/249).

TheAngryByrd (changes requested):
Left some feedback.

bartelink:
> I'd prefer not to make breaking changes here unless you want to retarget for https://github.com/demystifyfp/FsToolkit.ErrorHandling/pull/249

as mentioned in #254 I'm happy either way.

Adding it is not time-sensitive for me, so I'll leave it entirely up to you whether you want 
a) this PR based on v5
b) remove this diff and add a PR that can be merged later that does the contract change

I guess my slight preference is to stay with one PR, but I'm a maintainer too so I'd prefer you to make the call.

src/FsToolkit.ErrorHandling/Seq.fs
TheAngryByrd: Yeah either way I'm fine, since it's not time sensitive, lets retarget for v5.

paket.dependencies
bartelink: paket changes have been removed as they were pretty much there in v5

src/FsToolkit.ErrorHandling/Seq.fs
bartelink: rebased on v5

paket.dependencies
bartelink: re-did this after the rebase

bartelink: and then removed it - seems the lock file has many pending changes

build/build.fsproj
```diff
@@ -4,10 +4,12 @@
     <OutputType>Exe</OutputType>
     <TargetFramework>net7.0</TargetFramework>
     <IsPackable>false</IsPackable>
+    <!-- <NoWarn>NU1904</NoWarn>-->
+    <TreatWarningsAsErrors>false</TreatWarningsAsErrors>
```
bartelink, line 8: there's a new vulnerability that breaks the build
I see you have 1903 in the build props - if you fix that, I'l rebase off that as its obv not ideal to have this in the PR

bartelink:
The build wasn't happy for reasons I could not discern - happy to rebase or adjust this any time; just ping
By the same token, if you want to make adjustments and/or just merge it, that's fine too

PR #256, opened 2024-02-15
Seq sequence result a

### LINKED ISSUE #254: feat(Seq): Add sequenceResultA

bartelink:
WIP, nothing to see
includes changes from #255 

resolves #254

### DIFF
--- a/paket.dependencies
+++ b/paket.dependencies
@@ -1,18 +1,14 @@
 source https://api.nuget.org/v3/index.json
 
-
 storage: none
-
 nuget Microsoft.SourceLink.GitHub prerelease copy_local: true
-
-
 
 
 group NetStandard2
 source https://api.nuget.org/v3/index.json
 lowest_matching: true
 strategy: min
-nuget FSharp.Core >= 4.7.2
+nuget FSharp.Core >= 4.7.2 content: none
 nuget Ply
 nuget Hopac
 nuget FSharp.Control.AsyncSeq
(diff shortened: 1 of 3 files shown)


bartelink:
Abandoning, too much overlap with #255 so just putting two sets of changes into one in the interests of everyone's time

PR #261, opened 2024-04-06
209 remove MergeSources (and!) from some implementations

### LINKED ISSUE #209: Add asyncValidation CE, or make asyncResult (and result?) CEs work with 'and!'
I just discovered that I can't use `and!` in `result { }` and expect my errors to accumulate; I have to use the `validation { }`. However, I can't find a corresponding async variant. I have an `asyncResult` computation expression that uses `and!`, and I want the errors to accumulate. Is that possible, and if not, could you consider implementing it?

If it's at all possible, I think the best course of action would be to accumulate errors when using `and!` if the error type is `'a list`. This includes `result { }` and other similar CEs. Currently, `result { }` supports `and!` (i.e., the code compiles and runs), but there seems to be no difference between `and!` and successive `let!`, so it seems rather pointless. If `result { }` should not support applicative error accumulation, it would be better if `and!` was not supported at all; that would make it clear to users that they can't rely on applicative behavior, unlike now, where that is easily assumed but is not the case.

TheAngryByrd:
## Proposed Changes

Closes #209.  Removes MergeSources (so you can no longer `and!`) from:

- Result
- ResultOption
- AsyncResult
- Option
- ValueOption

I did not remove them from:

- TaskResult
- TaskOption

as those have some parallelism that still can be useful.

## Types of changes

## Checklist

## Further comments

If this is a relatively large or complex change, kick off the discussion by explaining why you chose the solution you did and what alternatives you considered, etc...


ursenzler:
Just wondering: is there a `parallelAsyncResult` that supports `and!`?

TheAngryByrd:
There is not but I’d take a PR for one!

PR #265, opened 2024-05-01
Update ce.md

m-rinaldi:
Added the type annotation to the argument of `tryParseInt` in order to fix the compile error:
> FS0041: A unique overload for method 'TryParse' could not be determined based on type information prior to this program point.


TheAngryByrd (approved):
Oh yeah, .NET has added lots of `ReadOnlySpan` overloads to many string accepting BCL members. I'm sure there's other places where this is affected too.


Thanks!

PR #274, opened 2024-07-21
Result.convertError convenience function added

m-rinaldi:
## Proposed Changes

This PR includes a convenience function, `Result.convertError`, intended to ease the work within a `result` computation expression with `Result` values whose error types are different but that can be converted to a broader error.

I've been writing this convenience function a few times, so I've finally decided to create a PR to include it.

### Example

Consider these two functions that return a `Result` value:

```
type NumberError = TooLow | TooHigh

let boundedNumber lower upper x =
  if x < lower then Error TooLow
  else if x > upper then Error TooHigh
  else Ok x


type StringError = TooShort | TooLong

let boundedString lower upper (s: string) =
  if s.Length < lower then Error TooShort
  else if s.Length > upper then Error TooLong
  else Ok s
```
The functions `boundedNumber` and `boundedString` can fail but with different errors: `NumberError` and `StringError`, respectively. Let's call these the *suberror* types.

Now, we define a new error type that can store the values of the two suberror types above, `NumberError` and `StringError`:

```
type BroaderError =
| NumErr of NumberError
| StrErr of StringError
```
We can attach the corresponding member functions that will make the conversion possible between the suberror types and the broader error type:

```
type BroaderError with
  static member From(err: NumberError) = NumErr err
  static member From(err: StringError) = StrErr err
```

Finally, it's possible to call `boundedNumber` and `boundedString` in the same computation expression and pipe their result to `Result.convertError`:

```
let overall: Result<_, BroaderError> = result {
  let! num = boundedNumber 7 113 11 |> Result.convertError
  and! str = boundedString 1 5 "foo" |> Result.convertError
  return num.ToString() + str
}
```
The type annotation `Result<_, BroaderError>` is needed to determine which `From` member to call. 

Without using the convenience function `Result.convertError`, we have to map the error values explicitly with `Result.mapError`:

```
let overall' = result {
  let! num = boundedNumber 7 113 11 |> Result.mapError (fun err -> BroaderError.From(err))
  and! str = boundedString 1 5 "foo" |> Result.mapError (fun err -> BroaderError.From(err))
  return num.ToString() + str
}
```

In this case, we don't need the type annotation, though.

## Types of changes

- [ ] Breaking change (fix or feature that would cause existing functionality not to work as expected)

## Checklist

### DIFF
--- a/gitbook/SUMMARY.md
+++ b/gitbook/SUMMARY.md
@@ -26,6 +26,7 @@
       * [traverseResultA](list/traverseResultA.md)
       * [sequenceResultA](list/sequenceResultA.md)
     * Transforms
+      * [convertError](result/convertError.md)
       * [ofChoice](result/ofChoice.md)
 
   * Option
--- a/gitbook/result/convertError.md
+++ b/gitbook/result/convertError.md
@@ -1,0 +1,83 @@
+# Result.convertError
+
+Namespace: `FsToolkit.ErrorHandling`
+
+## Function Signature
+
+```fsharp
+Result<'a, 'error1> -> Result<'a, ^error2>
+```
+
+`^error2` is a [statically resolved parameter](https://docs.microsoft.com/en-us/dotnet/fsharp/language-reference/generics/statically-resolved-type-parameters) with the below constraint:
+
+```fsharp
+^error2 : (static member From : 'error1 -> Result<'a, ^error2>)
+```
+
+This can be handy when working with result values whose types are parametrized with different error types if the overall result error value can be individually constructed from any of the intermediary error values. See the example below.
+
+## Example
+
+The following example is supposed to estimate the energy cost of climatizing a room. To do so, it must interact with a configuration file, temperature, humidity sensors, and a web service – any of these can potentially fail and come with custom error types.
+
+`readConfigFile` may fail with a `ConfigFileErr` error value:
+
+```fsharp
+type ConfigFileErr = FileNotFound | InsufficientPermissions
+
+let readConfigFile (filepath: string): Result<_, ConfigFileErr> =
+  Ok {|DesiredTemperature = 20.0; DesiredHumidity = 50.0|}
+```
+
+`getTemperature` and `getHumidity` may fail with error values of type `TempSensorErr` and `HumiditySensorErr`, respectively:
+
+```fsharp
+type TempSensorErr = Disconnected | Uncalibrated
+let getTemperature (): Result<_, TempSensorErr> = Ok 20.0
+
+type HumiditySensorErr = DeviceTimedOut
+let getHumidity (): Result<_, HumiditySensorErr> = Ok 60.0
+```
+
+`requestService` may fail with a `ServiceErr` error value:
+
+```fsharp
+type type ServiceErr = ServiceUnavailableErr
+
+open System
+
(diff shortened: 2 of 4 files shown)


brianrourkeboll:
Hmm.

If you want to go the `From` method route, why not just use a simple `convert` (or `upconvert` or `broaden` or whatever) function with `Result.mapError`?

```fsharp
let inline convert e = ((^error1 or ^error2) : (static member From : ^error1 -> ^error2) e)
```

That would let you do

```fsharp
let overall: Result<_, BroaderError> = result {
  let! num = boundedNumber 7 113 11 |> Result.mapError convert
  and! str = boundedString 1 5 "foo" |> Result.mapError convert
  return num.ToString() + str
}
```

I feel like this is too small of a diff to warrant the establishment of a whole new `From` method convention:

```diff
-Result.mapError convert
+Result.convertError
```

I would personally just use something like

```fsharp
let! … = result |> Result.mapError BroaderError.ofNarrow
and! … = …
```

or, keeping to the `From` convention,

```fsharp
let! … = result |> Result.mapError BroaderError.From
and! … = …
```

(You don't need the mapping lambda if the parameters line up — and if they _didn't_ line up, then you couldn't use `Result.convertError` anyway.)

Maybe someday we can get something like OCaml's polymorphic variants[^1][^2] in F#, and we can just use that instead :)

[^1]: https://ocaml.org/manual/5.2/polyvariant.html
[^2]: https://keleshev.com/composable-error-handling-in-ocaml

m-rinaldi:
> ```fsharp
> let inline convert e = (^error2 : (static member From : ^error1 -> ^error2) e)
> ```
> 
> That would let you do
> 
> ```fsharp
> let overall: Result<_, BroaderError> = result {
>   let! num = boundedNumber 7 113 11 |> Result.mapError convert
>   and! str = boundedString 1 5 "foo" |> Result.mapError convert
>   return num.ToString() + str
> }
> ```

Oh, I didn't know you could use SRTPs on functions you pass around.

> I would personally just use something like
> 
> ```fsharp
> let! … = result |> Result.mapError BroaderError.ofNarrow
> and! … = …
> ```
Right, it follows the existing naming conventions, e.g., `Option.ofObj`, `Option.ofNullable`, `List.ofArray`, etc.
> 
> (You don't need the mapping lambda if the parameters line up — and if they _didn't_ line up, then you couldn't use `Result.convertError` anyway.)

Do you mean thanks to the new [shorthand lambda syntax](https://devblogs.microsoft.com/dotnet/announcing-fsharp-8/#_-property-shorthand-for-fun-x-x-property)?

---

@brianrourkeboll Thank you, I will use your solution.

brianrourkeboll:
> Do you mean thanks to the new [shorthand lambda syntax](https://devblogs.microsoft.com/dotnet/announcing-fsharp-8/#_-property-shorthand-for-fun-x-x-property)?

Nope, I just meant normal [η-reduction](https://en.wikipedia.org/wiki/Lambda_calculus#%CE%B7-reduction) (eta-reduction).

I.e., given some function `f`, the expression `fun x -> f x` can (nearly) always be simplified to `f`.

So your original example of

```fsharp
boundedNumber 7 113 11 |> Result.mapError (fun err -> BroaderError.From(err))
```

can also be written

```fsharp
boundedNumber 7 113 11 |> Result.mapError BroaderError.From
```

m-rinaldi:
Got it now, thanks @brianrourkeboll

PR #279, opened 2024-10-22
Add Array errorhandling

DashieTM:
## Proposed Changes
This simply adds the functionality from lists to arrays by mirroring the functionality.
Alternative way would be to just always convert Arrays to and from lists.

## Types of changes

## Checklist

### DIFF
--- a/src/FsToolkit.ErrorHandling/Array.fs
+++ b/src/FsToolkit.ErrorHandling/Array.fs
@@ -1,0 +1,239 @@
+namespace FsToolkit.ErrorHandling
+
+[<RequireQualifiedAccess>]
+module Array =
+    let rec private traverseResultM' (state: Result<_, _>) (f: _ -> Result<_, _>) xs =
+        match xs with
+        | [||] ->
+            state
+            |> Result.map Array.rev
+        | arr ->
+            let x = Array.head arr
+            let xs = Array.skip 1 arr
+
+            let res =
+                result {
+                    let! y = f x
+                    let! ys = state
+                    return Array.append [| y |] ys
+                }
+
+            match res with
+            | Ok _ -> traverseResultM' res f xs
+            | Error _ -> res
+
+    let rec private traverseAsyncResultM'
+        (state: Async<Result<_, _>>)
+        (f: _ -> Async<Result<_, _>>)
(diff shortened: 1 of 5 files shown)


src/FsToolkit.ErrorHandling/Array.fs
```diff
@@ -0,0 +1,244 @@
...
+    let sequenceAsyncOptionM xs = traverseAsyncOptionM id xs
+
+#if !FABLE_COMPILER
+    let rec private traverseVOptionM' (state: voption<_>) (f: _ -> voption<_>) xs =
+        match xs with
+        | [||] ->
+            state
+            |> ValueOption.map Array.rev
+        | arr ->
+            let x = Array.head arr
```
bartelink, line 212: isnt there an array.split or similar that can do this pair of ops one shot?

src/FsToolkit.ErrorHandling/Array.fs
```diff
@@ -0,0 +1,244 @@
...
+            let xs = Array.skip 1 arr
+            let fR = f x
+
+            match state, fR with
+            | Ok ys, Ok y -> traverseValidationA' (Ok(Array.append [| y |] ys)) f xs
+            | Error errs1, Error errs2 ->
+                let errs =
+                    Array.concat [
```
bartelink, line 121: Array.append? (or at least don't allocate a `list` via `[]`?

src/FsToolkit.ErrorHandling/Array.fs
DashieTM: Not in a one shot, it returns 2 arrays, which is not what we want here.

src/FsToolkit.ErrorHandling/Array.fs
DashieTM: Thanks!

src/FsToolkit.ErrorHandling/Array.fs
bartelink: true, not sure what I was thinking!

DashieTM: All good, no worries :D

Issue #267, opened 2024-05-27
Combine use of Async.singleton & Async.retn


1eyewonder:
**Is your feature request related to a problem? Please describe.**
Currently, we have two functions within the Async module which do the same thing https://github.com/demystifyfp/FsToolkit.ErrorHandling/blob/master/src/FsToolkit.ErrorHandling/Async.fs#L6-L12

**Describe the solution you'd like**
Removing either  `Async.retn` or `Async.singleton` and replacing all references with the other would be the easiest solution. We might consider having a period where we convert and use the `[<Obsolete>]` attribute to announce the transition?

**Additional context**
Since this would be a breaking change, would this be worthwhile to add to the v5 changes coming soon?

PR #287, opened 2024-12-15
[v5] API Consolidation Around retn/singleton/ok/some

1eyewonder:
## Proposed Changes

Removed `Async.retn` since it was duplicating `Async.singleton`. This should simplify the API slightly. This addresses issue #267 

## Types of changes

- [X] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist

### DIFF
--- a/src/FsToolkit.ErrorHandling.IcedTasks/CancellableTaskValidationCE.fs
+++ b/src/FsToolkit.ErrorHandling.IcedTasks/CancellableTaskValidationCE.fs
@@ -324,12 +324,6 @@ module CancellableTaskValidation =
         fun _ -> Task.FromResult(x)
 
     /// <summary>Lifts an item to a CancellableTaskValidation.</summary>
-    /// <param name="item">The item to be the ok result of the CancellableTaskValidation.</param>
-    /// <returns>A CancellableTaskValidation with the item as the result.</returns>
-    let inline ok (item: 'ok) : CancellableTaskValidation<'ok, 'error> =
-        fun _ -> Task.FromResult(Ok item)
-
-    /// <summary>Lifts an item to a CancellableTaskValidation.</summary>
     /// <param name="error">The item to be the error result of the CancellableTaskValidation.</param>
     /// <returns>A CancellableTaskValidation with the item as the result.</returns>
     let inline error (error: 'error) : CancellableTaskValidation<'ok, 'error> =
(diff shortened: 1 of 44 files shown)


TheAngryByrd (changes requested):
Thanks for this! 

1. You should be able to rebase from master this so it doesn't include all those extra commits.
2. Should we remove all the other `retn` functions? Seems weird to only pick on `Async` module.

TheAngryByrd (changes requested):
Thanks for this! 

1. You should be able to rebase from master this so it doesn't include all those extra commits.
2. Should we remove all the other `retn` functions? Seems weird to only pick on `Async` module.

1eyewonder:
So I converted all uses of `retn`, `ok`, and `some` into singleton since they all do the same thing. For the cases with a Result inside, there are still the `error` functions. For naming consistency with `map`, `bind`, etc. do we think naming it to `singletonError` makes sense or did I overreach by changing the `ok` and `some` functions?

TheAngryByrd:
I think keeping `some` and `ok` would be better than forcing everyone to convert to `singleton` and especially `error` and `none` to `singletonError`.

1eyewonder:
Here are the current API changes so far. There may be some more massaging but hopefully this helps clarify the changes so far.

## IcedTasks API Changes

I'm guessing we may want to keep `ok` here to keep consistent with other validation-ish things

| Module                    | Notes                                    |
| ------------------------- | ---------------------------------------- |
| CancellableTaskValidation | Removed `ok`. Still contains `singleton` |

## JobResult API Changes

| Module          | Notes                           |
| --------------- | ------------------------------- |
| JobOption       | Converted `retn` to `singleton` |
| JobResult       | Converted `retn` to `singleton` |
| JobResultOption | Converted `retn` to `singleton` |

## Core Changes

For modules which contain both `singleton` & `some`/`ok` are we wanting to consolidate to just the latter?

| Module            | Notes                                               |
| ----------------- | --------------------------------------------------- |
| Async             | Removed `retn`. Still contains `singleton`          |
| AsyncOption       | Removed `retn`. Still contains `singleton` & `some` |
| AsyncResult       | Removed `retn`. Still contains `ok`                 |
| AsyncResultOption | Converted `retn` to `singleton`                     |
| AsyncValidation   | Removed `retn`. Still contains `ok`                 |
| ResultOption      | Converted `retn` to `singleton`                     |
| TaskOption        | Removed `retn`. Still contains `some`               |
| TaskResult        | Removed `retn`. Still contains `ok`                 |
| TaskResultOption  | Converted `retn` to `singleton`                     |
| Validation        | Removed `retn`. Still contains `ok`                 |

TheAngryByrd (approved):
This looks good! Thanks!

PR #288, opened 2024-12-15
[v5] Fantomas Updates

1eyewonder:
## Proposed Changes

1. Update fantomas to the latest version
2. Add `Main.fs` files from tests to .fantomasignore. The below code is causing errors with fantomas and thus not being caught. This also prevented us from catching code that needed formatting.

    ```fsharp
    #if !FABLE_COMPILER
    open Expecto
    
    [<Tests>] //needed for `dotnet test` to work
    #endif
    let allTests = testList "All Tests" [ AsyncSeq.allTests ]
    ```
3. `build.fs` has been updated to fail on uncaught code formatting errors within the `else` block.

## Types of changes

## Checklist

## Further Comments

I made it so that the `checkFormatCode` step fails if we encounter severe errors within Fantomas, however, I can also see why we may not want to fail here as well. I'm open to reverting/updating the `build.fs` as needed.

### DIFF
--- a/.config/dotnet-tools.json
+++ b/.config/dotnet-tools.json
@@ -1,30 +1,34 @@
 {
-    "version": 1,
-    "isRoot": true,
-    "tools": {
-        "paket": {
-            "version": "8.0.3",
-            "commands": [
-                "paket"
-            ]
-        },
-        "fable": {
-            "version": "4.4.0",
-            "commands": [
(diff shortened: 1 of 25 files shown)


1eyewonder:
After discussion on the issue thread above, I will make some quick updates later, so the code is not choking up fantomas

PR #290, opened 2024-12-15
[v5] Updated Uses of Seq.Append

1eyewonder:
## Proposed Changes

I was a silly goose and had remnants of `Seq.append` still in use within two different areas
1. `Seq.traverseResultA'`
2. `Seq.traverseAsyncResultA'`

Per discussions found in PR #277, we should be able to utilize a mixture of seq expressions and `Seq.rev` to provide better performance while also avoiding stack overflow issues.

## Types of changes

## Checklist

## Further Comments

I'd suggest looking/merging #288 first to simplify the git diff since fantomas & VSCode format on save updated this file

### DIFF
--- a/.config/dotnet-tools.json
+++ b/.config/dotnet-tools.json
@@ -1,30 +1,34 @@
 {
-    "version": 1,
-    "isRoot": true,
-    "tools": {
-        "paket": {
-            "version": "8.0.3",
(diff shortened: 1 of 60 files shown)


TheAngryByrd:
image


_Somehow_... `Seq.append` returned.

Issue #281, opened 2024-10-29
feat: Add 'Ensure' and 'Check' method similar to C#FunctionalExtensions


PI-Gorbo:
Hi! Loving the library. Using this library has felt very natural to me, as I use CSharpFunctionalExtensions all the time in C#. 
I think the library identifies some helpful methods that could be useful in this library too, and I have highlighted them below.
If there are already more 'FSharp' style ways of approaching this, I am very open. I don't mind being wrong :)

## [Ensure](https://github.com/vkhorikov/CSharpFunctionalExtensions/blob/master/CSharpFunctionalExtensions/Result/Methods/Extensions/Ensure.cs)
From the source code:
```
public static Result<T, E> Ensure<T, E>(this Result<T, E> result, Func<T, bool> predicate, Func<T, E> errorPredicate)
{
    if (result.IsFailure)
        return result;

    if (!predicate(result.Value))
        return Result.Failure<T, E>(errorPredicate(result.Value));

    return result;
}
```
Ensure allows you to evaluate a predicate. If the predicate is false, then return an error.

### Example usage of `ensure`
Before:
```
let mapIdentityResult (res: Task<IdentityResult>) : Task<Result<unit, string>> =
    res
    |> TaskResult.ofTask
    >>= fun res ->
            if res.Succeeded then
                TaskResult.ok ()
            else
                res.Errors
                    |> Seq.map (fun x -> x.Description)
                    |> Seq.fold (fun item agg -> agg + ", " + item) ""
                    |> TaskResult.error
```

After:
```
let mapIdentityResult (res: Task<IdentityResult>) : Task<Result<unit, string>> =
    res
    |> TaskResult.ofTask
    |> TaskResult.ensure (fun res -> res.Succeeded) (fun res ->
                    res.Errors
                    |> Seq.map (fun x -> x.Description)
                    |> String.concat ", ")
    |> TaskResult.ignore
```

## [Check](https://github.com/vkhorikov/CSharpFunctionalExtensions/blob/master/CSharpFunctionalExtensions/Result/Methods/Extensions/Check.cs)
From the source code:
```
public static Result<T, E> Check<T, K, E>(this Result<T, E> result, Func<T, Result<K, E>> func)
{
    return result.Bind(func).Map(_ => result.Value);
}
```
This method allows you to keep the calling success condition of the result.

### Example usage of Check
Here is a code snippet that could utilize check
```
...
  |> TaskResult.bind (fun user ->
[...]

PR #295, opened 2024-12-22
feat: Add Require and Check helper methods

PI-Gorbo:
## Context

Starting from this Task https://github.com/demystifyfp/FsToolkit.ErrorHandling/issues/281, I have implemented two new methods for Result, TaskResult and AsyncResult.

### Method 1 - Require
In the Issue i called it `ensure`, because that is what the C#FunctionalExtensions calls it, however I decided to name it `require` to be more in line with the other `requireX` functions.

Require applies a predicate to the Ok value of a Result. If the predicate is false, then a new Error Result is returned with a user provided error.

### Method 2 - Check
Same as described in the github task, check applies a more general function of the from ` 'ok -> Result<unit, 'error>` to the Ok value of a Result. It returns the original Ok value is the function returns Ok (), otherwise it returns the new error returned by the applied function.

I have added docs. 

I am super open to feedback and changes! Please fire away

### DIFF
--- a/gitbook/asyncResult/check.md
+++ b/gitbook/asyncResult/check.md
@@ -1,0 +1,76 @@
+# Result.check
+
+Namespace: `FsToolkit.ErrorHandling`
+
+The intent of check is to allow an Ok result value to be validated. 
+
+`check` takes a validation function of the form `'ok -> Async<Result<unit, 'error>>` and a result of the form `Async<Result<'ok, 'error>>`. 
+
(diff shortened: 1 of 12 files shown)


TheAngryByrd (commented):
Thanks for this! Sorry holidays have made reviewing hectic.

Ran the CI and got:

`[Expecto 00:00:18.7573892] FsToolkit.ErrorHandling.Tests: Found duplicated test names, these names are: [["All Tests.Result Tests.check tests.Using the result value in the predicate"]]` 

You'll have to disambiguate one of the tests.

PI-Gorbo:
Happy new year! Same with the holidays getting in my way.
I've sent a fix - I had a duplicated test. Sorry about that 🤦

TheAngryByrd (approved):
Thank you!

PR #310, opened 2025-02-17
feat(Seq.traverse/sequence*)!: Yield arrays

### LINKED ISSUE #254: feat(Seq): Add sequenceResultA

bartelink:
Instead of yielding a [lazy] sequence as the bodies of the Error or Ok state arising from the traverse and sequence results, yields Arrays directly. This offers two advantages:
- better pattern matching options - e.g. can match on `Ok [||]`, `Error [||]`, or use `_.Length` on the Error or Ok bodies (vs having to to `when Seq.length` etc)
- more efficient implementation (the current uses all sorts of temps via use of Seq.rev etc

Resolves #254

## Types of changes

In general, the signatures can be used interchangeably, but its undeniable that this is a binary breaking change.

## Checklist

### DIFF
--- a/src/FsToolkit.ErrorHandling/Seq.fs
+++ b/src/FsToolkit.ErrorHandling/Seq.fs
@@ -167,6 +167,7 @@ let traverseAsyncResultM f xs =
 /// <remarks>This function is equivalent to <see cref="traverseAsyncResultM"/> but auto-applying the 'id' function</remarks>
 let sequenceAsyncResultM xs = traverseAsyncResultM id xs
 
+#if !FABLE_COMPILER
 /// <summary>
 /// Applies a function to each element of a sequence and returns a single Task result
 /// </summary>
@@ -219,6 +220,8 @@ let traverseTaskResultM f xs =
 /// <returns>A task result with the ok elements in an array or the first error occurring in the sequence</returns>
 /// <remarks>This function is equivalent to <see cref="traverseTaskResultM"/> but auto-applying the 'id' function</remarks>
 let sequenceTaskResultM xs = traverseTaskResultM id xs
+
+#endif
 
 /// <summary>
 /// Applies a function to each element of a sequence and returns a single async result
--- a/src/FsToolkit.ErrorHandling/Seq.fs
+++ b/src/FsToolkit.ErrorHandling/Seq.fs
@@ -213,6 +213,14 @@ let traverseTaskResultM f xs =
     traverseTaskResultM' (TaskResult.ok Seq.empty) f xs
 
 /// <summary>
+/// Converts a sequence of Task results into a single Task result
+/// </summary>
+/// <param name="xs">The input sequence</param>
+/// <returns>A task result with the ok elements in an array or the first error occurring in the sequence</returns>
+/// <remarks>This function is equivalent to <see cref="traverseTaskResultM"/> but auto-applying the 'id' function</remarks>
+let sequenceTaskResultM xs = traverseTaskResultM id xs
+
+/// <summary>
 /// Applies a function to each element of a sequence and returns a single async result
 /// </summary>
 /// <param name="state">The initial state</param>
--- a/src/FsToolkit.ErrorHandling/Seq.fs
+++ b/src/FsToolkit.ErrorHandling/Seq.fs
@@ -1,5 +1,6 @@
-// See here for previous design disccusions:
+// See here for previous design discussions:
 // 1. https://github.com/demystifyfp/FsToolkit.ErrorHandling/pull/277
+// 2. https://github.com/demystifyfp/FsToolkit.ErrorHandling/pull/310
 
 [<RequireQualifiedAccess>]
 module FsToolkit.ErrorHandling.Seq
@@ -10,7 +11,7 @@ module FsToolkit.ErrorHandling.Seq
 /// <param name="state">The initial state</param>
 /// <param name="f">The function to apply to each element</param>
 /// <param name="xs">The input sequence</param>
-/// <returns>A result with the ok elements in an array or the first error occurring in the sequence</returns>
+/// <returns>If state and all elements in 'sequence' are Ok, a result with the elements in an array. Alternately, the first error encountered</returns>
 let inline traverseResultM'
     (state: Result<'okOutput seq, 'error>)
     ([<InlineIfLambda>] f: 'okInput -> Result<'okOutput, 'error>)
@@ -23,22 +24,20 @@ let inline traverseResultM'
     | Error e -> Error e
     | Ok initialSuccesses ->
 
-        use enumerator = xs.GetEnumerator()
-
-        let acc = ResizeArray(initialSuccesses)
-        let mutable err = Unchecked.defaultof<'error>
+        let oks = ResizeArray(initialSuccesses)
         let mutable ok = true
+        let mutable err = Unchecked.defaultof<'error>
         use e = xs.GetEnumerator()
 
         while ok
               && e.MoveNext() do
             match f e.Current with
-            | Ok r -> acc.Add r
+            | Ok r -> oks.Add r
             | Error e ->
                 ok <- false
                 err <- e
 
-        if ok then Ok(acc.ToArray()) else Error err
+        if ok then Ok(oks.ToArray()) else Error err
 
 /// <summary>
 /// Applies a function to each element of a sequence and returns a single result
@@ -53,7 +52,7 @@ let traverseResultM f xs = traverseResultM' (Ok Seq.empty) f xs
 /// Converts a sequence of results into a single result
 /// </summary>
 /// <param name="xs">The input sequence</param>
-/// <returns>A result with the ok elements in a sequence or the first error occurring in the sequence</returns>
+/// <returns>A result bearing all the results as an array, or the first error occurring in the sequence</returns>
 /// <remarks>This function is equivalent to <see cref="traverseResultM"/> but auto-applying the 'id' function</remarks>
 let sequenceResultM xs = traverseResultM id xs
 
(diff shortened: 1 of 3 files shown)


bartelink:
cc @1eyewonder @TheAngryByrd  My original impl was https://github.com/demystifyfp/FsToolkit.ErrorHandling/pull/255

I believe this correctly walks the sequences and combines successes and failures for the A variants (I reworded the docs as they suggested other semantics)

1eyewonder:
If you're looking to yield arrays directly, is there a reason why we are trying to implement in the Seq module as opposed to the existing [Array](https://github.com/demystifyfp/FsToolkit.ErrorHandling/blob/master/src/FsToolkit.ErrorHandling/Array.fs) module? The reason why I ended up converting this originally was due to needing lazy loading in very large sequences I was working with at the time. You can see the history shown here #277 if you didn't see this already. I'd also suggest adding benchmarks to the existing [benchmark tests](https://github.com/demystifyfp/FsToolkit.ErrorHandling/blob/master/benchmarks/SeqTests.fs) if we want to prove this is a more efficient implementation.

TheAngryByrd:
Sorry for the delay, I'll get to look at this in the weekend.

bartelink:
Not a problem - take as long as you need

bartelink:
Separate question - is there a plan/place for a Seq.traverseTaskResult ? I have a codebase that could use it and atm I have Seq.toList contortions

tests/FsToolkit.ErrorHandling.Tests/Seq.fs
```diff
@@ -72,13 +70,10 @@ let traverseOptionMTests =
                     "Hola"
                 }
 
-            let expected = Seq.toList tweets
+            let expected = Seq.toArray tweets
             let actual = Seq.traverseOptionM tryTweetOption tweets
 
-            let actual =
-                Expect.wantSome actual "Expected result to be Some"
-                |> Seq.toList
```
bartelink, line 80: note IMO needing to manipulate the result to look at it is a design smell (especially when there is no actual laziness, but yet you can't be 100% that you don't need to do a Seq.cache without seeing the impl as the type system does not guarantee it)

src/FsToolkit.ErrorHandling/Seq.fs
```diff
@@ -10,47 +11,48 @@ module FsToolkit.ErrorHandling.Seq
...
+        nullArg (nameof xs)
+
     match state with
-    | Error _ -> state
-    | Ok oks ->
-        use enumerator = xs.GetEnumerator()
+    | Error e -> Error e
+    | Ok initialSuccesses ->
+
+        let oks = ResizeArray(initialSuccesses)
```
TheAngryByrd, line 27: NIT: I think we could go with an [ArrayCollector](https://fsharp.github.io/fsharp-core-docs/reference/fsharp-core-compilerservices-arraycollector-1.html) to benefit any small enough sequences. Might as well squeeze out all the tiniest perf benefits.

TheAngryByrd (changes requested):
Since you seem to need to materialize the `seq` to know if it ends up being an `Ok` or an `Error`, it makes sense to just return the materialized array.

Only have a tiny nitpick, might as well use `ArrayCollector` and take care of any small perf benefits it could give.

@1eyewonder I think lazy consumption is still happening here, which is a desired trait. Does that seem like the case to you?

src/FsToolkit.ErrorHandling/Seq.fs
bartelink: Good point (my first time using it - it'd be nice to have a way to check for empty without the mutable flag abuse I added, though I suspect in terms of actual JITted IL this probably wins so not going to sweat it!

bartelink: That ArrayCollector shadowing behavior is probably something there should be a warnAsErrors setting or somesuch trapping?

bartelink:
Will also add the A variants, incoming...

bartelink:
Had to revert ArrayCollector stuff for stuff that goes through Fable (and the struct semantics forcing `mutable` are a bit confusing)

It would seem that the Array ones are now superfluous (or at least they should delegate to the Seq ones). My instinct is to take the former approach?

Similarly IMO the List ones are probably best off delegating to the Seq impls and mapping to List at the end, given the amount of allocations and temps involved in List.rev abuse in the current impls - ultimately this stuff all costs in compile/transpile/trim or execution time

bartelink:
Finally, a question re TFMs (I have form on this, see https://github.com/serilog/serilog/issues/1970)

In general for libs, I would try to obviously provide netstandard2.0 support where possible
net6.0 is also a useful TFM in that it has less messy dependency chains

However the incremental cost vs benefit of e.g. having a 9.0 one to my mind needs to be justified - unless you are actually calling net9.0-specific APs and/or otherwise benefitting from something new or better, the library itself should NOT be constantly updated to net8.0, net9.0, etc. It should instead either stick with `net6.0`, or not have a TFM-specific build

Obviously randomly dropping a TFM-specific package subset is not something to do lightly, as it can cause a mess, but it's a good idea if going to a new major to pick something minimal to support and stick with that

The test suites of course should follow a different policy - including at least testing with latest LTS, and optionally with the STS

bartelink:
And another spam - I dont know how to fix this with paket (and IMO paket is not a help for libs - the libs I maintain and/or co-maintain tend not to use paket, even if I use it for apps), but FSharp.Core versions prior to v7 (which I believe this lib should ideally reference) have xmldoc included that makes a mess of searching for things in Rider as it (correctly) shows the content. My workaround is: https://github.com/jet/equinox/blob/master/src/Equinox/Equinox.fsproj#L25-L28

bartelink:
@1eyewonder Note also my comments in https://github.com/demystifyfp/FsToolkit.ErrorHandling/pull/310#issuecomment-2692982877

I believe that the Array specific impls are now redundant (and I am pretty sure that the List ones should be implemented in terms of this. Ultimately while this is a binary breaking change, there are no cases where this signature is less useful for real usage.

TheAngryByrd:
> However the incremental cost vs benefit of e.g. having a 9.0 one to my mind needs to be justified - unless you are actually calling net9.0-specific APs and/or otherwise benefitting from something new or better, the library itself should NOT be constantly updated to net8.0, net9.0, etc. It should instead either stick with net6.0, or not have a TFM-specific build

I'm targeting TFM 9 because of F# Language Version 9 which brings in a lot of specific new nullable APIs. There was a lot of breaking Source overloads when I went up to LangVersion 9 so I'm constraining it to net9 only and having 8 be for lower. Also bumping language versions in F# can cause problems with consuming a library built with Lang9 and consuming from Lang8. See https://github.com/dotnet/fsharp/issues/14313 for an example of this.

TheAngryByrd:
> Let me know if you want me to do any of that; if not, I'm happy for this to get merged if you are...

I think this is fine for now. Yeah lets get this merged.

TheAngryByrd:
Thanks again for fixing this!

bartelink:
Thanks for getting it in - the release notes should probably more prominently stress that the return types have changed (i.e. I have the standard https://keepachangelog.com/en/1.0.0 `!` annotation implying the `BREAKING` you put in for some of the others. Any thoughts on alignment between Seq, Array and List impls? IMO

for Array, the impls are incomplete, and the desired signatures match the Seq ones, which means
  - add an Array facade that delegates to the Seq ones (but then would one sprout a ResizeArray or iarray equivalent if someone asked?
  - just remove it

While the List impl has a history and a track record, ultimately, for any interested number of results, `rev`ing a List and/or chaining a series of results in closures is just not a good idea (and the legibility difference with it being idiomatic recursive list manipulation does not outweigh that for me). However, there is a reasonable strong expectation of results being `list`. 

This suggests to me picking from:
- use the same impl, but replace `ResizeArray.ToArray` with `List.ofSeq` and `ArrayCollector.Close` with `ArrayCollector.Close >> List.ofSeq` (or implemented a CloseAsList, or migrate the temp to ResizeArray etc)
- add some internal helpers in the Seq module that enable them to be written as oneliners (i.e. parameterise the Close part to allow people to have arbitrary result types)

If you pop your outline lan/desires into an issue I guess it can be debated and/or closed from there...

TheAngryByrd:
> annotation implying the BREAKING you put in for some of the others. 

Yeah good point. I'll mark this.

> for Array, the impls are incomplete, and the desired signatures match the Seq ones, which means

Yeah I think it makes sense to delegate to the Seq ones, just to make discoverability better.


> se the same impl, but replace ResizeArray.ToArray with List.ofSeq and ArrayCollector.Close with ArrayCollector.Close >> List.ofSeq (or implemented a CloseAsList, or migrate the temp to ResizeArray etc)

For list specifically, there is a [ListCollector](https://github.com/dotnet/fsharp/blob/372d464ee2f4bfd09fd154c91a6b3be8b1218f3e/src/FSharp.Core/seqcore.fs#L535) that's very much in the same vein as ArrayCollector. If Fable doesn't support it, might have to `#if`.

Issue #228, opened 2023-09-25
Support TaskValidation to match AsyncValidation


AlexeyRaga:
We have `AsyncValidation`, it would be logical to have a matching `TaskValidation` functionality.

**Describe alternatives you've considered**
As an alternative, I could stay with `AsyncValidation` and then do `Async.StartAsTask`, but it is less straightforward.

TheAngryByrd:
I agree. Would you mind submitting a pull request for it?

AlexeyRaga:
Maybe, but not promising, not quickly at very least :(

tw0po1nt:
@TheAngryByrd  looks like this can be closed. I guess I didn't use the proper secret word for Github to automatically code this when the referencing PR was merged

PR #313, opened 2025-03-01
Add `TaskValidation` module

tw0po1nt:
## Proposed Changes

Add `TaskValidation` module as proposed in #228 

## Types of changes

## Checklist

## Further comments

This will probably warrant a decent amount of scrutiny. The `TaskValidation` module mirrors the `AsyncValidation` module, along with its suite of tests, same with the ops. The code for the computation expression and its tests are derived from the `TaskResult` computation expression and its tests, since the `AsyncValidation` computation expression does not use the state machine. This should probably be fine, since `Validation` is simply a `Result` that accumulates the errors. But it may warrant some additional, `Validation` specific tests that I might have overlooked.

### DIFF
--- a/src/FsToolkit.ErrorHandling/TaskValidationCE.fs
+++ b/src/FsToolkit.ErrorHandling/TaskValidationCE.fs
@@ -1,0 +1,556 @@
+namespace FsToolkit.ErrorHandling
+
+open System.Threading.Tasks
+open System.Runtime.CompilerServices
+open System.Threading
+open Microsoft.FSharp.Core
+open Microsoft.FSharp.Core.CompilerServices
+open Microsoft.FSharp.Core.CompilerServices.StateMachineHelpers
+open Microsoft.FSharp.Core.LanguagePrimitives.IntrinsicOperators
+open Microsoft.FSharp.Control
+open Microsoft.FSharp.Collections
+
+[<Struct; NoComparison; NoEquality>]
+type TaskValidationStateMachineData<'T, 'Error> =
+
+    [<DefaultValue(false)>]
+    val mutable Validation: Validation<'T, 'Error>
+
+    [<DefaultValue(false)>]
+    val mutable MethodBuilder: AsyncTaskValidationMethodBuilder<'T, 'Error>
+
+    member this.IsValidationError = Result.isError this.Validation
+    member this.IsTaskCompleted = this.MethodBuilder.Task.IsCompleted
+
+and AsyncTaskValidationMethodBuilder<'TOverall, 'Error> =
+    AsyncTaskMethodBuilder<Validation<'TOverall, 'Error>>
+
+and TaskValidationStateMachine<'TOverall, 'Error> =
+    ResumableStateMachine<TaskValidationStateMachineData<'TOverall, 'Error>>
+
+and TaskValidationResumptionFunc<'TOverall, 'Error> =
+    ResumptionFunc<TaskValidationStateMachineData<'TOverall, 'Error>>
+
+and TaskValidationResumptionDynamicInfo<'TOverall, 'Error> =
+    ResumptionDynamicInfo<TaskValidationStateMachineData<'TOverall, 'Error>>
+
+and TaskValidationCode<'TOverall, 'Error, 'T> =
+    ResumableCode<TaskValidationStateMachineData<'TOverall, 'Error>, 'T>
+
+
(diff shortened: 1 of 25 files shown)


tests/FsToolkit.ErrorHandling.Tests/FsToolkit.ErrorHandling.Tests.fsproj
```diff
@@ -34,9 +34,16 @@
         <Compile Include="ValidationCE.fs" />
         <Compile Include="AsyncValidation.fs" />
         <Compile Include="AsyncValidationCE.fs" />
+        <Compile Include="TaskValidation.fs" />
+        <Compile Include="TaskValidationCE.fs" />
         <Compile Include="Main.fs" />
         <None Include="paket.references" />
         <Watch Include="@(None)" />
     </ItemGroup>
+    <ItemGroup>
```
tw0po1nt, line 43: @TheAngryByrd this doesn't seem like the right approach for this...? I updated the package via Rider because I needed the `testCaseTask` function. The `TaskResultCE` test file references a version of Expecto with this function but doesn't do this.

tw0po1nt: indeed, the build checks are failing because of this.

tw0po1nt: Fixed by moving the tests into the `FsToolkit.ErrorHandling.TaskResult.Tests` project

tw0po1nt:
TODO: Documentation

tw0po1nt (commented):
self-review complete

tests/FsToolkit.ErrorHandling.TaskResult.Tests/TaskValidationCE.fs
```diff
@@ -0,0 +1,740 @@
...
+                        let! a = Choice1Of2 3
+                        and! b = Choice1Of2 2
+                        and! c = Choice2Of2 errorMsg
+                        return a + b - c
+                    }
+
+                Expect.equal actual (Validation.error errorMsg) "Should be Error"
+            }
+
+        testCaseTask "Fail Path Validation/Choice/AsyncValidation"
```
TheAngryByrd, line 709: Unless I don't see it, I don't think there's tests covering seeing multiple failure cases.  Something like how https://github.com/demystifyfp/FsToolkit.ErrorHandling/blob/def0948c21d5716122d776861c55ebfcfc5bd82d/tests/FsToolkit.ErrorHandling.Tests/ValidationCE.fs#L448 is working.

src/FsToolkit.ErrorHandling/TaskValidationCE.fs
```diff
@@ -0,0 +1,577 @@
...
+            t
+            |> Task.map Ok
+
+        member inline this.Source(computation: Async<'T>) : TaskValidation<'T, 'Error> =
+            computation
+            |> Async.map Ok
+            |> Async.StartImmediateAsTask
+
+[<AutoOpen>]
+module TaskValidationCEExtensionsHighPriority2 =
```
TheAngryByrd, line 558: I think this is missing overloads for `Result<'a,'err>` (where `'err` is not a list).  You can see examples of the overload resolution in AsyncValidation: https://github.com/demystifyfp/FsToolkit.ErrorHandling/blob/def0948c21d5716122d776861c55ebfcfc5bd82d/src/FsToolkit.ErrorHandling/AsyncValidationCE.fs#L126

TheAngryByrd (changes requested):
Lots of great work here!

I left some comments on areas I think are missing.

tests/FsToolkit.ErrorHandling.TaskResult.Tests/TaskValidationCE.fs
tw0po1nt: Added a test covering that scenario in latest commit. Let me know if that is what you were looking for.

src/FsToolkit.ErrorHandling/TaskValidationCE.fs
tw0po1nt: Added these overloads

TheAngryByrd (approved):
❤️ Thank you!

PR #327, opened 2025-05-09
Add ok and error helper functions to TaskResultOption and AsyncResultOption modules

tw0po1nt:
## Proposed Changes

Add a couple of simple but useful helper functions to modules with the `Result` type in their stack

## Types of changes

## Checklist


src/FsToolkit.ErrorHandling/AsyncResultOption.fs
```diff
@@ -24,6 +24,13 @@ module AsyncResultOption =
             )
             input
 
+    let inline ok x =
```
1eyewonder, line 27: Nit: maybe we want to start adding some xml doc comments to these functions with the links to where they will be found like what we have been adding recently.

1eyewonder (commented):
Overall looks good! Maybe we also want to add `ok` to the ResultOption as well. Here an old PR that came to mind when looking at this for context #287

src/FsToolkit.ErrorHandling/AsyncResultOption.fs
tw0po1nt: Sure, I can do that. I didn't here because the others didn't and I wondered if it wasn't because the function is so simple

1eyewonder:
Nope, that actually makes perfect sense. I'm ok leaving it out. One less thing we have to maintain 😄

TheAngryByrd (approved):

Issue #319, opened 2025-03-13
Add `taskValueOption` and `valueTaskValueOption` CEs


xperiandri:
**Is your feature request related to a problem?**
`taskOption` CE exists but `taskValueOption` does not

**Describe the solution you'd like**
Add `taskValueOption` and `valueTaskValueOption` CEs

PR #329, opened 2025-05-31
Add TaskValueOption module, operators, and CE + tests and documentation

tw0po1nt:
## Proposed Changes

Added some code requested in #319 

## Types of changes

## Checklist


TheAngryByrd:
I did some updates around the build infrastructure in #331 so this might fix that build error if you can rebase

tw0po1nt:
@TheAngryByrd I rebased...looks like even more checks are failing now. Looks like it's complaining about nullness:

```sh
TaskValueOptionCE.fs(241,33): error FS3264: Nullness warning: Downcasting from 'objnull' into 'ICriticalNotifyCompletion' can introduce unexpected null values. Cast to 'ICriticalNotifyCompletion|null' instead or handle the null before downcasting.
```

It's strange, because it's not failing with that locally, either building via `build.sh` or through Rider.

tw0po1nt:
Looks like I just needed to pull in some changes from the previous PR and apply them here. Should be good to go now.

PR #345, opened 2025-11-15
Net10

1eyewonder:
## Proposed Changes

- Upgrade from net9.0 -> net10.0 in relevant areas.
- Turning on analyzers in the IDE from the recent switch from paket to nuget
- Some general versioning updates to keep health of the repo up to date

## Types of changes

## Checklist

## Further comments

If this is a relatively large or complex change, kick off the discussion by explaining why you chose the solution you did and what alternatives you considered, etc...

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -19,12 +19,12 @@
         },
         // https://github.com/devcontainers/features/blob/main/src/github-cli/README.md
         "ghcr.io/devcontainers/features/github-cli:1": {},
-        // https://github.com/devcontainers-contrib/features/blob/main/src/starship/README.md
-        "ghcr.io/devcontainers-contrib/features/starship:1": {},
+        // https://github.com/devcontainers-extra/features/blob/main/src/starship/README.md
+        "ghcr.io/devcontainers-extra/features/starship:1": {},
         // https://github.com/devcontainers/features/blob/main/src/dotnet/README.md
         "ghcr.io/devcontainers/features/dotnet:2": {
-            "version": "9.0",
-            "additionalVersions": "8.0"
+            "version": "10.0",
+            "additionalVersions": ["8.0", "9.0"]
         },
         "ghcr.io/devcontainers/features/node:1": {
             "version": "22"
--- a/.config/dotnet-tools.json
+++ b/.config/dotnet-tools.json
@@ -3,14 +3,14 @@
   "isRoot": true,
   "tools": {
     "fable": {
-      "version": "4.25.0",
+      "version": "4.27.0",
       "commands": [
         "fable"
       ],
       "rollForward": false
     },
     "fantomas": {
-      "version": "7.0.2",
+      "version": "7.0.3",
       "commands": [
         "fantomas"
       ],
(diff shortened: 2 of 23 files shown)


.devcontainer/devcontainer.json
```diff
@@ -19,12 +19,12 @@
...
-        // https://github.com/devcontainers-contrib/features/blob/main/src/starship/README.md
-        "ghcr.io/devcontainers-contrib/features/starship:1": {},
+        // https://github.com/devcontainers-extra/features/blob/main/src/starship/README.md
+        "ghcr.io/devcontainers-extra/features/starship:1": {},
         // https://github.com/devcontainers/features/blob/main/src/dotnet/README.md
         "ghcr.io/devcontainers/features/dotnet:2": {
-            "version": "9.0",
-            "additionalVersions": "8.0"
+            "version": "10.0",
+            "additionalVersions": ["8.0", "9.0"]
```
1eyewonder: I don't think this is necessary but temporarily have it here

TheAngryByrd (changes requested):
I appreciate the work but we shouldn't replace net9.0 -> net10.0.

Things we should not do:

- Replace net9.0 outright. 
    - The 9 sdk is still under support until next year https://dotnet.microsoft.com/en-us/platform/support/policy/dotnet-core
- We shouldn't replace tfm net9.0 -> net10.0 in libraries (things under src)
    - This would be a breaking change
    - The reason you choose a TFM in libraries is because it brings a set of APIs you need.
        - In the case of netstandard2.1, it's because it provides IAsyncEnumerable/IAsyncDisposable/ValueTask natively
        - In the case of net6.0, it's because it provides [PooliingValueTask](https://learn.microsoft.com/en-us/dotnet/api/system.runtime.compilerservices.poolingasyncvaluetaskmethodbuilder-1?view=net-6.0) support 
        - In the case of 9.0, it's because of the [Nullable/Nullness](https://devblogs.microsoft.com/dotnet/nullable-reference-types-in-fsharp-9/) support in F# 9.0. (This is technically not a TFM limitation but an FSharp.Core version limitation. I didn't want to start forcing people to use FSharp.Core 9 on older TFMs so I drew a line in the sand at TFM 9.0)
        - The only reason we would want to build against 10.0 is the [RuntimeAsync](https://github.com/dotnet/runtime/blob/main/docs/design/specs/runtime-async.md) work. However, there's an [F# compiler limitation](https://github.com/dotnet/fsharp/issues/19056) at the moment, so we can't ship with this until that's fixed. 
            - possibly tail recursion in the form of [ReturnFromFinal/YieldFromFinal](https://learn.microsoft.com/en-us/dotnet/fsharp/whats-new/fsharp-10#tail-call-support-in-computation-expressions). But this would be a similar thing as Nullness support, we'd want TFM 10 + FSharp.Core 10 but leave all others as less.
    - In either case, it would be additive and not remove the build target of net9.0


Things we can do:

- Add (not replace) net10.0 to test projects if we want to see how they work against 10 runtime
- Add net10.0 to benchmarks
- Update the various infrastructure like docker/yml/global.json to support being able to run 10.

bartelink:
Has this been superseded? #373 is bringing similar stuff too...

1eyewonder:
@bartelink Yes apologies. I've been busy with other things so if you're continuing work we can use your branch and I can close this one out

bartelink:
No need to apologize - you've done lots of great work in the lib; was just checking we weren't in conflict on the off chance...

PR #359, opened 2026-04-13
Add partitionResults with single-pass destructuring

### LINKED ISSUE #358: Question/Feature: partition collections of results into ok cases and error cases.
When dealing with collections of Results it can in some cases be useful partition Ok cases from Error cases. For instance when doing something that can error partially during batch processing or similar I have found myself having to work with a `Result<'A, 'B>[]` where it would make a lot of sense to return a data structure indicating "Partial Success" and including the Ok cases and Error cases separately. 

Therefore, I would find it very useful to have a partition function which also destructures the result types such that I can deal with both ok cases and error cases. I have solved this using this implementation which I thought might be nice to have in this project. The below implementation should of course be extended to other collections as well.
```fsharp
[<RequireQualifiedAccess>]
module Array =
    let partitionResults (input: Result<'a,'b>[]) : 'a[] * 'b[] =                                                                                                                                               
      input |> Array.choose (function Ok v -> Some v | _ -> None),                                                                                                                                            
      input |> Array.choose (function Error e -> Some e | _ -> None) 
```

Perhaps there is a smarter way of doing this, and if so, I would love to be enlightened.

bpe-incom:
## Proposed Changes

Adds `partitionResults` functions to `Array`, `List`, and `Seq` modules that partition a collection of `Result<'ok, 'error>` into a tuple of unwrapped ok values and error values.

Closes #358

## Implementation

All implementations use single-pass traversal as suggested in the issue discussion:
- **Array**: Uses `ResizeArray` accumulators (per @njlr's suggestion)
- **List**: Uses tail-recursive loop with cons/rev (idiomatic for F# lists)
- **Seq**: Uses `ResizeArray` accumulators, returns `seq` rather than `array`

### DIFF
--- a/gitbook/SUMMARY.md
+++ b/gitbook/SUMMARY.md
@@ -26,11 +26,13 @@
       * [sequenceResultM](list/sequenceResultM.md)
       * [traverseResultA](list/traverseResultA.md)
       * [sequenceResultA](list/sequenceResultA.md)
+      * [partitionResults](list/partitionResults.md)
     * Sequences
       * [traverseResultM](seq/traverseResultM.md)
       * [sequenceResultM](seq/sequenceResultM.md)
       * [traverseResultA](seq/traverseResultA.md)
       * [sequenceResultA](seq/sequenceResultA.md)
+      * [partitionResults](seq/partitionResults.md)
     * Transforms
       * [ofChoice](result/ofChoice.md)
 
--- a/gitbook/SUMMARY.md
+++ b/gitbook/SUMMARY.md
@@ -33,6 +33,8 @@
       * [traverseResultA](seq/traverseResultA.md)
       * [sequenceResultA](seq/sequenceResultA.md)
       * [partitionResults](seq/partitionResults.md)
+    * Arrays
+      * [partitionResults](array/partitionResults.md)
     * Transforms
       * [ofChoice](result/ofChoice.md)
 
--- a/src/FsToolkit.ErrorHandling/Array.fs
+++ b/src/FsToolkit.ErrorHandling/Array.fs
@@ -235,6 +235,9 @@ module Array =
 #endif
 
     let partitionResults (input: Result<'ok, 'error>[]) : 'ok[] * 'error[] =
+        if isNull input then
+            nullArg (nameof input)
+
         let oks = ResizeArray()
         let errors = ResizeArray()
 
--- a/src/FsToolkit.ErrorHandling/Array.fs
+++ b/src/FsToolkit.ErrorHandling/Array.fs
@@ -235,5 +235,12 @@ module Array =
 #endif
 
     let partitionResults (input: Result<'ok, 'error>[]) : 'ok[] * 'error[] =
-        input |> Array.choose (function Ok v -> Some v | _ -> None),
-        input |> Array.choose (function Error e -> Some e | _ -> None)
+        let oks = ResizeArray()
+        let errors = ResizeArray()
+
+        for x in input do
+            match x with
+            | Ok v -> oks.Add v
+            | Error e -> errors.Add e
+
+        oks.ToArray(), errors.ToArray()
--- a/src/FsToolkit.ErrorHandling/Array.fs
+++ b/src/FsToolkit.ErrorHandling/Array.fs
@@ -233,3 +233,7 @@ module Array =
     let sequenceVOptionM xs = traverseVOptionM id xs
 
 #endif
+
+    let partitionResults (input: Result<'ok, 'error>[]) : 'ok[] * 'error[] =
+        input |> Array.choose (function Ok v -> Some v | _ -> None),
+        input |> Array.choose (function Error e -> Some e | _ -> None)
--- a/src/FsToolkit.ErrorHandling/Array.fs
+++ b/src/FsToolkit.ErrorHandling/Array.fs
@@ -234,8 +234,13 @@ module Array =
 
 #endif
 
+    /// <summary>
+    /// Partitions an array of results into a tuple of the ok values and the error values.
+    /// </summary>
+    /// <param name="input">The input array of results.</param>
+    /// <returns>A tuple where the first element is an array of all Ok values and the second is an array of all Error values.</returns>
     let partitionResults (input: Result<'ok, 'error>[]) : 'ok[] * 'error[] =
-        if isNull input then
+        if System.Object.ReferenceEquals(input, null) then
             nullArg (nameof input)
 
         let oks = ResizeArray()
--- a/src/FsToolkit.ErrorHandling/List.fs
+++ b/src/FsToolkit.ErrorHandling/List.fs
@@ -276,6 +276,9 @@ module List =
 #endif
 
     let partitionResults (input: Result<'ok, 'error> list) : 'ok list * 'error list =
+        if isNull input then
+            nullArg (nameof input)
+
         let oks = ResizeArray()
         let errors = ResizeArray()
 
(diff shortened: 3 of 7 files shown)


src/FsToolkit.ErrorHandling/Seq.fs
```diff
@@ -501,3 +501,22 @@ let traverseVOptionM f xs =
 let sequenceVOptionM xs = traverseVOptionM id xs
 
 #endif
+
+/// <summary>
+/// Partitions a sequence of results into a tuple of the ok values and the error values.
+/// </summary>
+/// <param name="xs">The input sequence of results.</param>
+/// <returns>A tuple where the first element is an array of all Ok values and the second is an array of all Error values.</returns>
+let partitionResults (xs: SeqNull<Result<'ok, 'error>>) : 'ok seq * 'error seq =
```
bartelink: there is prior art in this repo for the Seq variant yielding arrays. There is no real benefit to downcasting here as it's hard to conceive of a different internal impl that beats accumulating and materializing

So my suggestion here would be for Seq.X to delegate to and Array.X to share this impl. On the array side, let the inputs be Arrays so it could be optimized to e.g. consider the input lengths. On the Seq side you can't but returning arrays lets people pattern match and/or lenght check the arrays immediately vs pretending its lazy and/or them upcasting again

src/FsToolkit.ErrorHandling/List.fs
```diff
@@ -274,3 +274,12 @@ module List =
 
 
 #endif
+
+    let partitionResults (input: Result<'ok, 'error> list) : 'ok list * 'error list =
+        let rec loop oks errors =
+            function
+            | [] -> List.rev oks, List.rev errors
```
bartelink: idiomatic or not, this is pretty horrible in terms of allocations - better to loop, accumulating into a ResizeArray and yield via List.ofSeq ?

For my first pass impl, I'd even consider implementing it in terms of the generic Seq/Array one, i.e. it could be a 1/2 liner: `let partitionResults xs = let oks, errs = Seq.partitionResults xs in List.ofArray oks, List.ofArrays errs`

src/FsToolkit.ErrorHandling/Seq.fs
bartelink: found it: https://github.com/demystifyfp/FsToolkit.ErrorHandling/pull/310
one notable thing is that ArrayCollector can be useful in this space but in the context of this lib stick with ResizeArray so it can be supported in the Fable impl (which did not have ArrayCollector at least at that time)

src/FsToolkit.ErrorHandling/List.fs
bpe-incom: I tried with a new approach which doesn't condense the logic into a one-liner, but implements your first suggestion.

src/FsToolkit.ErrorHandling/Array.fs
```diff
@@ -233,3 +233,14 @@ module Array =
     let sequenceVOptionM xs = traverseVOptionM id xs
 
 #endif
+
+    let partitionResults (input: Result<'ok, 'error>[]) : 'ok[] * 'error[] =
+        let oks = ResizeArray()
```
bartelink, line 246: probably add a null guard on the input, not sure if that's been universally applied in other such functions.

For List, its less necessary, but it is nonetheless useful for a lib like this to have a guard (e.g. for reflection or serialization cases where people wind up with a `null` instance, even if the language precludes doing it directly.

other than that, the signatures LGTM :shipit:  - kudos on getting in and contributing!

src/FsToolkit.ErrorHandling/List.fs
bartelink: 👍 yes, it's all tradeoffs, and for cases like this when a simple algorithm isnt logn and is easily scannable, composing  lots of cryptic pieces often doesnt beat something you can eyeball in an instant ;)

src/FsToolkit.ErrorHandling/Array.fs
bpe-incom: Makes sense. I added null guards to array as well as the list implementation for good measure.

src/FsToolkit.ErrorHandling/Seq.fs
```diff
@@ -501,3 +501,22 @@ let traverseVOptionM f xs =
 let sequenceVOptionM xs = traverseVOptionM id xs
 
 #endif
+
+/// <summary>
+/// Partitions a sequence of results into a tuple of the ok values and the error values.
+/// </summary>
+/// <param name="xs">The input sequence of results.</param>
+/// <returns>A tuple where the first element is an array of all Ok values and the second is an array of all Error values.</returns>
+let partitionResults (xs: SeqNull<Result<'ok, 'error>>) : 'ok[] * 'error[] =
```
bartelink: sorry for late breaking pickiness... probably rename xs to input for consistency with the others (and add xmldoc for them)

src/FsToolkit.ErrorHandling/Array.fs
```diff
@@ -233,3 +233,17 @@ module Array =
     let sequenceVOptionM xs = traverseVOptionM id xs
 
 #endif
+
+    let partitionResults (input: Result<'ok, 'error>[]) : 'ok[] * 'error[] =
```
TheAngryByrd, line 242: Nit: XML docs

src/FsToolkit.ErrorHandling/List.fs
```diff
@@ -274,3 +274,17 @@ module List =
 
 
 #endif
+
+    let partitionResults (input: Result<'ok, 'error> list) : 'ok list * 'error list =
```
TheAngryByrd, line 283: Nit: XML docs

TheAngryByrd (changes requested):
Thanks for this! Looks good with two outstanding things:

1. Couple nitpicks for XML docs
2. Add corresponding docs over in https://github.com/demystifyfp/FsToolkit.ErrorHandling/tree/master/gitbook

(also thanks for all the review work @bartelink!)

src/FsToolkit.ErrorHandling/Array.fs
TheAngryByrd: I'm not too concerned about nullArg on `list` types, reflection deserializing have a million footguns and you could check everywhere all the time. But there's a reason I made: 

https://github.com/demystifyfp/FsToolkit.ErrorHandling/blob/a4a153bf24798437d2899ae20fd900a5c9879f4e/src/FsToolkit.ErrorHandling/Option.fs#L208-L224.

I'm fine with either no check or using `System.Object.ReferenceEquals`

gitbook/SUMMARY.md
```diff
@@ -26,11 +26,13 @@
       * [sequenceResultM](list/sequenceResultM.md)
       * [traverseResultA](list/traverseResultA.md)
       * [sequenceResultA](list/sequenceResultA.md)
+      * [partitionResults](list/partitionResults.md)
     * Sequences
       * [traverseResultM](seq/traverseResultM.md)
       * [sequenceResultM](seq/sequenceResultM.md)
       * [traverseResultA](seq/traverseResultA.md)
       * [sequenceResultA](seq/sequenceResultA.md)
+      * [partitionResults](seq/partitionResults.md)
```
bartelink, line 35: don't see array ref'd here - seed a new section if required...

bpe-incom: Yeah, we could do that, but it is already just repeats of the same docs for lists and seq where an array section simply would be another permutation.

As a user of the library, I have never 'missed' the array documentation. But we can add it in this commit if @TheAngryByrd would find it useful?

TheAngryByrd: Yeah create a new section please. Don't have to backfill anything that's missing, just reference anything that already exists.

TheAngryByrd (approved):
Thank you for the work!

PR #360, opened 2026-04-21
Fix exceptions handling in task dynamic invocation

auduchinok:
Fixes task handling when dynamic invocations are used.

The context is I'm working on https://github.com/dotnet/fsharp/pull/19548 which improves debugging experience by preventing inlining where possible and allowing stepping into inline functions and setting breakpoints there. This makes the compiler produce the dynamic invocations instead of the state machines, and this uncovered exceptions in FsToolkit.ErrorHandling tests.


TheAngryByrd (approved):
Thank you!

PR #373, opened 2026-09-11
Rebase impl, docs and surface API on FSharp.Core v 11 APIs

### LINKED ISSUE #154: Task to AsyncResult transition wraps `Exception` into the `AggregateException`
**Describe the bug**
```
    member this.AsyncAcquireTokenInteractively (uiThread, correlationId) = asyncResult {
        do! Async.SwitchToContext uiThread
        let builder =
            pca.AcquireTokenInteractive(options.Scopes)
                .WithCorrelationId(correlationId)
        if not (isNull configureInteractive) then configureInteractive.Invoke(builder)
        try return! builder.ExecuteAsync() // FSharp.Control.FusionTasks is used not to write |> Async.AwaitTask
        with :? MsalClientException as ex -> return! Error ex.ErrorCode
    }
```
Matching by the `MsalClientException` does not happen as it is wrapped by the `AggregateException`.

**To Reproduce**
1. Call Task returning method via `return!` in an `asyncResult` CE
2. Catch exception
3. You will see that it is always the `AggregateException`.

**Expected behavior**

**Desktop (please complete the following information):**
 - Version 2.10.0

### LINKED ISSUE #361: Provide catching alternative to `AsyncResult.ofAsync`?
This code will throw an exception:

```fsharp
#r "nuget: FsToolkit.ErrorHandling, 5.2.0"

open FsToolkit.ErrorHandling

let workflow =
  async {
    failwith "kaboom"

    return 1
  }

asyncResult {
  let! x =
    workflow
    |> AsyncResult.ofAsync

  return x + 1
}
|> Async.RunSynchronously
|> printfn "%A"
```

But users might want to catch and print:

```
Error
  System.Exception: kaboom
```

The following behaves with catching behavior:

```fsharp
#r "nuget: FsToolkit.ErrorHandling, 5.2.0"

open FsToolkit.ErrorHandling

let workflow =
  async {
    failwith "kaboom"

    return 1
  }

asyncResult {
  let! x =
    workflow
    |> AsyncResult.ofAsync
    |> AsyncResult.catch id

  return x + 1
}
|> Async.RunSynchronously
|> printfn "%A"
```

Perhaps a catching helper should be provided?

```fsharp
[<RequireQualifiedAccess>]
module AsyncResult =

  let catchAsync (workflow : Async<'a>) : Async<Result<'a, exn>> =
    async {
      try
        let! res = workflow

        return Ok res
      with ex ->
        return Error ex
    }
```

This would be opposite of `AsyncResult.getOrReraise`. 

Can send a PR. Thanks!

### LINKED ISSUE #372: Add/use shim versions of FSharp.Core Async and Task modules?
Now that https://github.com/fsharp/fslang-suggestions/issues/1466 https://github.com/fsharp/fslang-suggestions/issues/1467 have made it to `fsharp/main`, I'm wondering if it might make sense to put poor man's versions of the impls in as shims and then have net11.0 builds use the in the box ones?

Thinking e.g. having clones of `Async.Await`, `Task.parallelLimit` in the box conditionally compiled in for <= `net11.0` and making the libs internal calls use `Async.Await`, `Async.catch` and friends to the max degree possible (which would fix longstanding issues on this lib yielding `AggregateException` etc.

Or is there a point where the lib simply takes a hard net11 / net12 dependency? (Yes I could search, but more trying to get a quick take from you - I'm not literally as bored as the question might make me seem!)

bartelink:
FSharp.Core adds utility functions for Async, Task, ValueTask as covered in https://github.com/fsharp/fslang-suggestions/issues/1466 and https://github.com/fsharp/fslang-suggestions/issues/1467 that overlap significantly / borrow from FsToolkit.Error Handling.

In the short term, there are some signature clashes (Task.catch in FSharp.Core 11 returns Result, not Choice)
In the medium term, this repo won't yet adopt SDK 11, and the net9.0 and earlier builds won't have access to the new Async.Await.

In the long term that stuff gets removed from here and removed from the docs

Thus this PR:
- adds a net10 TFM build that:
  - references FSharp.Core 11 (-rc1, which is reasonable given v6 here on master is in beta)
  - has signatures consistent with the long term names (remove Task.catch to avoid the clash)
  - has names consistent with the FSharp.Core names
    - removes Task/Async/others.singleton (in favor ofTask.result), Result.valueOr (Result.defaultWith)
      - v6 will probably do a breaking change to rename singleton to result for consistency
    - removes things that shadow FSharp.Core intrinsics (e.g. Result.map,mapError, bind, defaultValue, defaultWith, isOk, isError,) 
- does shimming and minor changes for net9/net6/Fable
  - adds as many shims as possible for e.g. Async.Await, Task.*, Async.*

NOTE The FSharp.Core shims re Async and Task match the behavior (pass the test suite) and surface area with the following deviations:
- `Async.Await does` not fully replicated the nested stack trace preservation behavior that FSharp.Core v7 provides (top level stack trace is shown, but the full trace during a trampolined nested call can only be provided by the real implementation). (Obviously it does the unwrapping demanded by #154 etc)
- the TaskLike Async.Wait versions (and related StartTaskImmediate helpers) are excluded due to compilation issues in the tests (you end up disambiguating via `Async.Await` as `Async.Await<_>` to exclude a clash with `Async.Await<_, _, _` ) which does not present with the real FSharp.Core's API surface).

Resolves #372
Resolves #154
Resolves #361


gitbook/SUMMARY.md
```diff
@@ -361,7 +361,7 @@
     * [apply](job/apply.md)
     * [map2](job/map2.md)
     * [map3](job/map3.md)
-    * [singleton](job/singleton.md)
+    * [result](job/result.md)
```
bartelink: we could just remove this as this is the builtin one from Hopac

src/FsToolkit.ErrorHandling.JobResult/JobOption.fs
```diff
@@ -20,10 +20,17 @@ module JobOption =
             return! t
         }
 
-    let inline singleton x = Job.result (Some x)
+    let inline some x = Job.result (Some x)
+
+    [<System.Obsolete "Use JobOption.some instead (aligns with AsyncResult naming)">]
+    let inline singleton x = some x
+
+    //TODO should these be added alongside all `some` ?
```
bartelink, line 28: thoughts?

src/FsToolkit.ErrorHandling/Async.fs
```diff
@@ -127,31 +128,15 @@ module Async =
...
                 map box input2
             |]
         )
-#else
-        Async.Parallel(
-            [|
-                map box input1
-                map box input2
-            |],
-            maxDegreeOfParallelism = 2
```
bartelink, line 143: does not achieve anything

bartelink:
@TheAngryByrd still WIP but worth a review now:
- Ended up putting full shims for new FSharp.Core 11 Task and Async modules (but not ValueTask)
- `Await` and usage of same in favor of `AwaitTask` is still todo. Arguably that should be a separate PR
- ignore should probably gain `RequiresExplicitTypeArguments` for Job etc to align with Task/Async
- should the Option variants have a `none` alongside `some`? (see self-review comment)
- anything else (I don't have energy/capacity/spare credits to expand docs but think I've tracked all renames) ?
- arguably docs should be removed for Task/Async functions that move to FSharp Core and Job.result (intrinsic to Hopac). See _(from FSharp.Core)_ annotations in summary.md

TheAngryByrd:
> adds a net10 TFM build that: references FSharp.Core 11

I think that we should consider going to `net11`, there may be a need to do `net10` with FSharp.Core 10 in the future.

bartelink:
> I think that we should consider going to `net11`, there may be a need to do `net10` with FSharp.Core 10 in the future.

That was my initial instinct; I only changed tack due to it being a preview SDK and the CI not being rigged for it as yet.

Definitely don't want to release a non-beta release depending on FSharp.Core 11-rc

That said, non-LTS TFM-specific builds have less value for me and should be candidates for trimming (leave net6.0 etc, but remove net9.0, and maybe run the tests under net8.0 (to cover the net8.0 TFM build)).

I'd venture running tests under net9.0 should also removed, i.e. maybe replace the existing net9.0 stuff with net10.0 as part of this release cycle then we pin its API surface via Package Validation (e.g. https://github.com/jet/equinox/blob/master/src/Equinox/Equinox.fsproj#L5).

So maybe:
1. leave net6.0 + net8 build/test
2. add package validation for net6 + net10 builds
3. when we have an RC or final net11 we want to do a real -rc with:
  - provision sdk in CI, update global.json
  - change net10.0 tfm-build to net11.0, test under net11
  - change net9.0 tfm-build to net10.0, test under net10

bartelink:
@TheAngryByrd This one goes after #376 - no point reviewing while the diffs combine both. Will rebase once you merge the other. I have no plans to implement anything else in the context of this API (though maybe [`RunSynchronouslyImmediate`](https://github.com/dotnet/core/blob/main/release-notes/11.0/preview/rc1/fsharp.md#asyncrunsynchronouslyimmediate)) can be done if I get around to it.

TheAngryByrd:
> > I think that we should consider going to `net11`, there may be a need to do `net10` with FSharp.Core 10 in the future.
> 
> That was my initial instinct; I only changed tack due to it being a preview SDK and the CI not being rigged for it as yet.
> 
> Definitely don't want to release a non-beta release depending on FSharp.Core 11-rc


Yeah makes sense

> That said, non-LTS TFM-specific builds have less value for me and should be candidates for trimming (leave net6.0 etc, but remove net9.0, and maybe run the tests under net8.0 (to cover the net8.0 TFM build)).
> 
>
> I'd venture running tests under net9.0 should also removed, i.e. maybe replace the existing net9.0 stuff with net10.0 as part of this release cycle then we pin its API surface via Package Validation (e.g. https://github.com/jet/equinox/blob/master/src/Equinox/Equinox.fsproj#L5)

Yeah, eventually we'll want to just remove net8/net9 `Exe` tests, but I'm still hesitant to remove it at the Api/Library level. _The curse_ of being a semi-popular library is you don't know how many people are stuck on old runtimes (like game devs). While we don't test against older runtimes, we only add APIs, we don't have drastically different implementations of each functions on each runtime so validating with tests on newer runtimes is fine.

1. Remove net8 and net9 tests
2. Add net11 test
3. global.json to net11

bartelink:
So in the short term, a beta with net10 TFMs built with SDK 10 as standin ?

And then flip as above on net11 release (net10 -> net11 for the SDK and TFM-specific builds of prod builds).

When we reach net11, we keep test running on net10 (to exercise net9.0 TFM specific builds).

Not sure it makes sense to keep running an unsupported net8 env at at that point (though IcedTasks has a net6.0 - should that be removed?)

Maybe you could transplant your interpretation/answers to this into an issue for the net11 release timeframe?

bartelink:
@TheAngryByrd  FYI this is the shims in context showing the poor man's Await passes the full FSharp.Core test suite (except for the 5 about the nested trampolied ones)
https://github.com/bartelink/fsharp/commit/3dd2045c334413a243be4bc6b78a566107ea7974

